PR #1, opened 2018-04-12
Zah review changes

coffeepots:


yglukhov:
There's one design issue that bothers me, and with this pr it becomes apparent. The issue is that eth procs are implied to be implemented in a single (rigidly included) file. This is because the `rpc` macro is so tempting, but it is not available to the rest of the world. TBH I don't consider the `rpc` macro useful at all (it's really simplistic and does straightforward but not obvious transformations), instead I would prefer smth like the following:

```nim
# rpc_debug_interface.nim
import rpc_server
# ... more imports

let s = sharedRpcServer()

s.on("debug.verbosity") do(a: JsonNode) -> JsonNode:
  # change verbosity

s.on("debug.dropDb") do(a: JsonNode):
  # drop db
```
```nim
# rpc_eth_interface.nim
import rpc_server

let s = sharedRpcServer()
s.on("eth.getBallance") do(a: JsonNode) -> JsonNode:
  # get balance
```
Basically your impl already allows something similar, it's just about the decoupling approach. Further improvements could be auto marshalling of Jsons:
```nim
s.on("debug.dropDb") do() -> bool:
  # drop db. Note how I'm not taking any args, and returning bool.
  return true

s.on("debug.verbosity") do(level: int):
  # change verbosity. Note how I'm taking int arg, and returning nothing.
```

Hope that makes sense.

coffeepots:
@yglukhov 
Before this PR, the ethprocs were a separate semi-independent module, but it became apparent that in order to get the 'built in' ethproc rpcs into the compile time list (that automatically sets up new servers with procs registered using the `{.rpc.}` macro), and also allow 3rd party rpc procs too, the ethprocs would have to be visible at compile time to the initServer call.

This is because, as far as I can tell, once a module is imported it's macros are expanded and thus the code is 'fixed' to only have the calls that have already been processed by `{.rpc.}` at the module import time.

Honestly, I'm not happy about this set up. Working in included files is not a great experience; you lose all your type checking for that file as, at least in this case, the server types aren't visible to ethprocs.

I had hoped to have ethprocs as a (theoretically) optional import that just added the default Ethereum rpc calls, but as it stands it's not this flexible.

In the current scheme, if the registration is taken out of `newRpcServer`, ethprocs wouldn't need to be included, but this forces the user to have to manually call `registerRpcs(server)` even to get the 'built in' ethprocs calls, which I had wanted to avoid.

Your macro ideas are interesting and something to consider. I agree that the `{.rpc.}` macro isn't necessarily giving you gains when writing rpc calls, especially as with this PR it constrains ethprocs to require being included in the main server types module.

I like the idea of namespaces and being able to separate out calls from VM to debug and other situations. I also like the `do:` variant of writing calls and marshalling to JSON automatically.

However I do feel that decoupling is sorely needed at this stage.

@zah do you have any thoughts on this?

yglukhov:
@coffeepots 

>> in order to get the 'built in' ethproc rpcs into the compile time list

Is there such a necessity? I can't see any other usages of `rpcCallRefs` except adding into it, and passing it to `register` in runtime.

>> In the current scheme, if the registration is taken out of newRpcServer, ethprocs wouldn't need to be included, but this forces the user to have to manually call registerRpcs(server) even to get the 'built in' ethprocs calls, which I had wanted to avoid.

So with my approach there would be no need to additionally register anything. The `on` function would just add the handler to the server.

>> However I do feel that decoupling is sorely needed at this stage.

I'm pretty sure the limitations will be visible very soon. `servertypes` includes `ethprocs`. Now correct me if I misunderstood your intention. `servertypes` itself sounds like a pretty common name that is going to end up being imported in a lot of other `nimbus` modules (directly or indirectly). It is likely you're going to import nimbus files in `ethprocs` (which is bad on its own, because that would introduce cyclic package dependency), and it is extremely likely to introduce import cycles. Breaking the cycles after they live for a while in a project is a big pita from my experience :).

coffeepots:
> It is likely you're going to import nimbus files in ethprocs (which is bad on its own, because that would introduce cyclic package dependency), and it is extremely likely to introduce import cycles.

I think this is reason for a refactor on its own. I'm going to look at this today.

yglukhov:
>> I believe so, as otherwise you need to re-define your rpc calls for each server you set up. However there's probably a better way to do this.

Adding the same set of handlers to different servers (if that's what you mean) is pretty easy with my suggestion:
```nim
proc addDebugFunctions(s: RpcServer) =
  s.on("debug.dropDB") do():
    # drop db
  s.on("debug.verbosity") do(level: int):
    # set verbosity

addDebugFunctions(sharedRpcServer())
addDebugFunctions(anotherRpcServer())
```

zah:
@yglukhov, one detail that you might be missing is that I requested that these handlers should be async procs (because replying to some RPC call may require executing other async operations).

This can be done with your proposed interface, but it will have a transformation.

Regarding the use of an included file, we should kill this indeed. We don't need to use compile-time mechanisms for everything. It's perfectly fine for the macro to call a run-time `registerRpcHandler` proc, which just adds a function pointer to some run-time table. It woudn't even be a problem to spread the implementations across multiple files then.

yglukhov:
> one detail that you might be missing is that I requested that these handlers should be async procs

@zah, yes, I skipped those intentionally to simplify reading of my snippets. The `do` closures should return `Future[JsonNode]` at the most generic level. But with automarshalling it could be done automagically for sync procs.

PR #5, opened 2018-04-24
Remove includes and expand macros

coffeepots:
I've updated the interface for the server so it's easier to create RPC calls.

Can I get both of your feedback on the interface?

The `ethprocs` have been liberated from their `include` prison.
You can now use the following to create an RPC call.

```
s = sharedRpcServer() # this is currently just a hidden lazy-init global
s.on("the.path5") do(b: string, s: seq[int], arr: array[20, byte]):
  # Json values are fetched into the above native Nim types for use
```

There are some more examples in `serverTypes.nim` and `testserverclient.nim`.
I've also started converting `ethprocs` over to use the new interface but I need to implement object marshalling to keep some of the rpc calls in the same interface style.

I'm envisioning being able to pass object definitions as parameters like you can currently with other types, and have them initialised from the json - potentially with errors raised/transmitted for you if the types/fields aren't what's expected.

I left in the `echo result.repr` at the end of the `on` macro so you can see how the procs end up.

There is still some work to do:

* Return values are not marshalled yet
* Object values require marshalling
* Paths need more work - currently they're just a straight up string compare
* Probably more error checking
* Errors need to be async return values, not exceptions (otherwise every wrong parameter call will crash the server)
* General code tidying and prettifying


eth-rpc/client/clientdispatch.nim
```diff
@@ -6,53 +6,46 @@ type
...
   
-  # REVIEW: These shouldn't be just asserts. You cannot count
-  # that the other side implements the protocol correctly, so
-  # you must perform validation even in release builds.
-  assert node.hasKey("jsonrpc")
-  assert node["jsonrpc"].str == "2.0"
-  assert node.hasKey("id")
-  assert self.awaiting.hasKey(node["id"].str)
+  # TODO: Use more appropriate exception objects
+  if not node.hasKey("jsonrpc"): raise newException(ValueError, "Message is missing rpc version field")
```
yglukhov: Minor optimisation suggestion: `hasKey("key")` and `node["key"]` both do lookups. Number of lookups can be reduced to 1 with `node{"key"}`.

eth-rpc/server/ethprocs.nim
```diff
@@ -1,188 +1,184 @@
-import servertypes, json, asyncdispatch, rpcregistration
-import cryptoutils
+import servertypes, cryptoutils, json, macros
```
yglukhov: This file will likely have to be moved to nimbus repo.

tests/testrpcmacro.nim
```diff
@@ -0,0 +1,105 @@
...
+
+var s = newRpcServer("localhost")
+
+# RPC definitions
+s.on("rpc.simplepath"):
+  echo "hello3"
+  result = %1
+
+s.on("rpc.returnint") do() -> int:
+  echo "hello2"
```
coffeepots: I'll remove this one later, as I have the return tests below it.

eth-rpc/server/servertypes.nim
```diff
@@ -13,15 +14,145 @@ type
...
+  let emptyStr = ident"eStr"
+  for item in values:
+    # generate tuples of values with the empty string `eStr`
+    let sItem = $item
+    multiReplaceCall.add(newPar(newStrLitNode(sItem), emptyStr))
+
+  body.add multiReplaceCall
+  result = newBlockStmt(body)
+
+macro on*(server: var RpcServer, path: string, body: untyped): untyped =
```
yglukhov: How likely is the pattern
```nim
s.on("path"):
  doSmth
```
vs
```nim
s.on("path") do(someArg: int) -> SomeResult:
  doSmth
```
If `do` is more likely, maybe it is not worth maintaining the other branch, and user's code will be more consistent as well =)
I think it would be beneficial to make `body` typed and expect `nkDo`, or `nkProcDef`, or friends. Then `parameters` would be just `body.params`, so it should be somewhat easier, and more strict as to what the user can feed to your macro.

You can always play with `typed` later, but only if you prohibit the anything-but-do way now. ;)

eth-rpc/server/servertypes.nim
```diff
@@ -13,15 +14,145 @@ type
...
+proc unpackArg[T](argIdx: int, argName: string, argtype: typedesc[T], args: JsonNode): T =
+  fromJson(args[argIdx], argName, result)
+
+proc setupParams(parameters, paramsIdent: NimNode): NimNode =
+  # Add code to verify input and load parameters into Nim types
+  result = newStmtList()
+  if not parameters.isNil:
+    # initial parameter array length check
+    var expectedLen = parameters.len - 1
+    let expectedStr = "Expected " & $expectedLen & " Json parameter(s) but got "
```
yglukhov: I'd better use a callout to some `validateParameters(expectedCount, jNode)`. This simplifies the macro, and keeps bin size smaller.

PR #7, opened 2018-05-08
PR #5 (Remove includes and expand macros) change requests

coffeepots:
* Moved raises out of generics (need to replace these with proper async returns)
* Refactored check code for expected parameter length
* Moved `multiRemove` macro into proc
* Used accessors to ensure checking when fetch required message fields from json


eth-rpc/server/servertypes.nim
```diff
@@ -13,15 +14,152 @@ type
...
+proc unpackArg[T](argIdx: int, argName: string, argtype: typedesc[T], args: JsonNode): T =
+  when argType is array or argType is seq:
+    if args[argIdx].kind != JArray: raise newException(ValueError, "Parameter \"" & argName & "\" expected JArray but got " & $args[argIdx].kind)
+  when argType is array:
+    if args[argIdx].len > result.len: raise newException(ValueError, "Parameter \"" & argName & "\" item count is too big for array")
+  when argType is object:
+    if args[argIdx].kind != JObject: raise newException(ValueError, "Parameter \"" & argName & "\" expected JObject but got " & $args[argIdx].kind)
+  fromJson(args[argIdx], argName, result)
+
+proc expectArrayLen(node: NimNode, paramsIdent: untyped, length: int) =
```
yglukhov: Sorry, what i meant by this is not a compile-time proc, but a simple non-generic runtime proc, and the `setupParams` would just add a call to it, if that makes sense. Sorry for the confusion. Anyways, this is by no means critical.

coffeepots: @yglukhov I've replaced the macro now :)

eth-rpc/server/serverdispatch.nim
```diff
@@ -20,7 +20,7 @@ proc processMessage(server: RpcServer, client: AsyncSocket, line: string) {.asyn
     if not server.procs.hasKey(methodName):
       await client.sendError(METHOD_NOT_FOUND, "Method not found", id, %(methodName & " is not a registered method."))
     else:
-      let callRes = server.procs[methodName](node["params"])
+      let callRes = await server.procs[methodName](node["params"])
```
zah: If a bogus JSON is received, the dictionary and field accesses here will raise an exception, right? Is the rest of the code prepared for this? Perhaps we should try to enumerate the possible exceptions returned by the JSON validation and marshaling code.

eth-rpc/server/servertypes.nim
```diff
@@ -13,15 +14,141 @@ type
...
+    # and pass it back jsonified - of course, we don't want to do this
+    # if a JsonNode is explicitly declared as the return type     
+    let
+      returnType = parameters[0]
+      res = ident"result"
+    template doMain(body: untyped): untyped =
+      # create a new scope to allow shadowing result
+      block:
+        body
+    bodyWrapper = quote do:
```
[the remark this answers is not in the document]
coffeepots: No, it does not :(

eth-rpc/server/jsonconverters.nim
```diff
@@ -0,0 +1,24 @@
+import json, stint
+
+iterator bytes(i: UInt256|Int256): byte =
+  let b = cast[ptr array[32, byte]](i.unsafeaddr)
```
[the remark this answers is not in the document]
coffeepots: Agree with this.

eth-rpc/client/clientdispatch.nim
```diff
@@ -77,37 +82,17 @@ proc connect*(self: RpcClient, address: string, port: Port) {.async.} =
...
-      call = newCall(newDotExpr(ident("client"), ident("call")), newStrLitNode(callName), ident("params"))
-      body = newStmtList().add call
-      templ = makeTemplate(callName, params, body, true)
-    params.add newNimNode(nnkBracketExpr).add(ident("Future"), ident("Response"))
-    params.appendFormalParam("client", "RpcClient")
-    params.appendFormalParam("params", "JsonNode")
-    result.add templ
+    let nameLit = ident(callName)
+    result.add(quote do:
+      template `nameLit`*(client: RpcClient, params: JsonNode): Future[Response] = client.call(`callName`, params)  # TODO: Back to template
```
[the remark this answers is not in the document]
coffeepots: I've updated `ethprocs` to use type signatures, so although the root RPC is accepting a `JsonNode`, the `on` macro processes this into Nim types.

All the RPCs accept Json arrays for parameters, and the spec says the RPC should accept a Json array for it's parameters. As such we should only need one `JsonNode` as a parameter, but the `on` macro allows clearer type specifications.
For example: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getbalance

I can upload the updated `ethprocs`, but unfortunately I have a bug to fix in `servertypes.nim` somewhere, so it's complaining about not being able to find `result` - in the past this has actually been because it couldn't process a parameter type, and I think currently this is caused by something returning a non type when the macro is expecting an expression.

eth-rpc/server/servertypes.nim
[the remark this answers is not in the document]
coffeepots: I presume I would have to do as `{.async.}` does and, IIRC, search through for the `return` keyword and replace it with the shadowed result and exit the block to the handler.

eth-rpc/server/serverdispatch.nim
coffeepots: I agree we need to enumerate exceptions and reduce them to the smallest quantity possible.

However, there's a check for the method name key just above it so this shouldn't cause a problem, should it?

eth-rpc/client/clientdispatch.nim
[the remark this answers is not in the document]
coffeepots: @zah see the updated `ethprocs` I just uploaded.

eth-rpc/server/jsonconverters.nim
```diff
@@ -0,0 +1,24 @@
...
+  var pos = 0
+  while pos < 32:
+    yield b[pos]
+    pos += 1
+
+proc `%`*(n: UInt256): JsonNode =
```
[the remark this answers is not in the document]
coffeepots: Yes I've been wondering over using string or a constrained length array for many of the types.

eth-rpc/client/clientdispatch.nim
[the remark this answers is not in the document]
coffeepots: I see what you mean about the client procs though. Would be nice to have them specified as the server ones are.

eth-rpc/server/servertypes.nim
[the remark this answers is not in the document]
zah: You can just put the body in a separate async proc and `await` its result here.

eth-rpc/server/jsonconverters.nim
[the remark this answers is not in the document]
coffeepots: UInts are now processed as strings with a length check.

eth-rpc/server/serverdispatch.nim
coffeepots: Extra checking added.

eth-rpc/client/clientdispatch.nim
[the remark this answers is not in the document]
coffeepots: Clients can now use native nim calls for ethprocs, which are generated from the signatures in `ethsigs.nim`.
These automatically marshal data from Nim to json types, perform the call, then marshal any return types from json back into native Nim types.

eth-rpc/client/clientdispatch.nim
```diff
@@ -77,37 +89,102 @@ proc connect*(self: RpcClient, address: string, port: Port) {.async.} =
...
+    callBody.add(quote do:
+      `procRes` = `rpcResult`.result
+      )
+  when defined(nimDumpRpcs):
+    echo pathStr, ":\n", result.repr
+
+from os import getCurrentDir, DirSep
+from strutils import rsplit
+
+macro processRpcSigs(): untyped =
```
zah: If this is going to become a library, this macro can take a file path in the form of identifiers:

```nim
jsonRcpProcotol ../my_protocol
```
Then the NimNode tree of operators and identifiers can be converted to a path with `$`.

eth-rpc/client/clientdispatch.nim
```diff
@@ -77,37 +89,102 @@ proc connect*(self: RpcClient, address: string, port: Port) {.async.} =
...
+  let
+    rpcResult = genSym(nskLet, "res") # temporary variable to hold `Response` from rpc call
+    procRes = ident"result"           # proc return variable
+    jsonRpcResult =                   # actual return value, `rpcResult`.result
+      nnkDotExpr.newTree(rpcResult, newIdentNode("result"))
+  
+  # perform rpc call
+  callBody.add(quote do:
+    let `rpcResult` = await client.call(`pathStr`, `jsonParamIdent`)          # `rpcResult` is of type `Response`
+    if `rpcResult`.error: raise newException(ValueError, $`rpcResult`.result) # TODO: is raise suitable here?
```
zah: I think an exception here is appropriate, but you should define your own exception types instead of using `ValueError` everywhere.

The exception will be put in the `Future` object representing the call. There might be a short-cut for doing this without raising an exception. If there isn't such a shortcut now, @cheatfate can add one to the new `asyncdispatch2` lib.

eth-rpc/client/clientdispatch.nim
```diff
@@ -77,37 +89,102 @@ proc connect*(self: RpcClient, address: string, port: Port) {.async.} =
...
+proc toJsonArray(parameters: NimNode): NimNode =
+  # outputs an array of jsonified parameters
+  # ie; %[%a, %b, %c]
+  parameters.expectKind nnkFormalParams
+  var items = newNimNode(nnkBracket)
+  for i in 2 ..< parameters.len:
+    let curParam = parameters[i][0]
+    if curParam.kind != nnkEmpty:
+      items.add(nnkPrefix.newTree(ident"%", curParam))
+  result = nnkPrefix.newTree(newIdentNode("%"), items)
```
zah: You can try to use `bindSym("%", brForceOpen)` here instead of using `newIdentNode`.

rpcserver.nim
```diff
@@ -0,0 +1,7 @@
+import
+  eth-rpc / server / servertypes,
```
zah: why do avoid using the idiomatic syntax for module imports?

```nim
import
  eth-rpc/server/[servertypes, ehtprocs, etc]
```

eth-rpc/client/clientdispatch.nim
coffeepots: Just updated my local branch to do just that :)

rpcserver.nim
coffeepots: I had some problems with this actually, which is why it ended up like that.
Using
```
import
  eth-rpc/server/[servertypes, rpcconsts, serverdispatch, ethprocs]
```
I get:
`cannot open file: eth-rpc/server/`, where the error points to the hyphen.
Then a second error, which is probably just as a result of the first error messing up parsing:
`invalid module name: '[servertypes, rpcconsts, serverdispatch, ethprocs]`

I've had a second look, and I can get around the issue by doing this, though it looks kind of odd:
```
import
  r"eth-rpc/server" / [servertypes, rpcconsts, serverdispatch, ethprocs]
```
Seems as if the dash is messing things up, but only if you're using path / [moduleA, moduleB] syntax.

zah: I guess the best approach is to avoid dashes in the directory names in general.

PR #12, opened 2018-06-01
Remove requirement for passing server when using rpc macro

coffeepots:
This PR was made after discussions with @cheatfate about the `rpc` macro's dependency on passing an `RpcServer` object around in order to automatically register it.

The rational is that by passing a server to bind to, you're potentially breaking the separation of concerns in submodules which shouldn't need to know how the server side is set up, or worse, requiring a global simply so that you can satisfy the automatic `server.register` call inside `rpc`. This inhibits writing thread safe code and obstructs modularisation of RPCs to different server instances.

RPCs registered to multiple servers allows:

* Multiple RPC servers on a single thread, or use multiple threads without worrying about a dependency on globals or threadvars.
* Balancing workload for high CPU or RPCs called with high frequency across different internal servers, allowing smarter management with limited resources.

To address this, I've split `rpc` into three macros:

* The original `rpc` is now called `registerRpc` to explicitly show it is doing the extra work of registering an RPC to a particular server.
* `rpc` that simply creates the procedure without registering
* `rpc` that creates the procedure with a generated name, and passes this name back to you so you can register it with any path you wish, without worrying about proc name clashes.

Additionally, to help support these new mechanisms, `register` has been updated to have three variants:

* Register using vararg path names
    `srv.register("a.b", "b.c")`
* Register using a symbol
    `srv.register(procName)`
* Register using an array of symbols
    `srv.register([ab, bc, d])`

This is currently a work in progress and will breaks tests for now.
I've added some examples at the bottom of `servertypes.nim` to show how the calls are used.

### DIFF
--- a/eth-rpc/server/servertypes.nim
+++ b/eth-rpc/server/servertypes.nim
@@ -14,10 +14,7 @@ type
     code*: int
     data*: JsonNode
 
-proc register*(server: RpcServer, name: string, rpc: RpcProc) =
-  server.procs[name] = rpc
-
-proc unRegisterAll*(server: RpcServer) = server.procs.clear
+var sharedServer: RpcServer
 
 proc newRpcServer*(address = "localhost", port: Port = Port(8545)): RpcServer =
   result = RpcServer(
@@ -26,8 +23,6 @@ proc newRpcServer*(address = "localhost", port: Port = Port(8545)): RpcServer =
     address: address,
     procs: newTable[string, RpcProc]()
   )
-
-var sharedServer: RpcServer
 
 proc sharedRpcServer*(): RpcServer =
   if sharedServer.isNil: sharedServer = newRpcServer("")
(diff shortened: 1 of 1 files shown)


coffeepots:
This PR has already been incorporated in @cheatfate's PR #13, but is easier to see the macro directly here.

zah:
I don't understand the motivation here. What's wrong with having the RCP procs registered on a particular server instance? You can again create as many server instances as you want and you can create helper procs that register a particular set of procs to a particular server.

This API makes you jump through hoops for the common case of having one server handling all of the RPC procs.

coffeepots:
@zah 
You can create multiple server instances, but you can only do so after you've attached it to an `RpcServer` instance.
So, in order to register an RPC to mutliple servers, you'd have to do this:

```nim
passedRpcServerInstance.rpc("myrpc"):
  # ...
anotherRpcServer.add("myrpc", rpcMyRpc)
```

As such, you have to have a global passed around because the RPC procs are exported, so you can't do this for example:
```nim
proc addRpcSubset*(server: RpcServer) =
  server.rpc("rpc1"):
    ## ...
  server.rpc("rpc2"):
    ## ...
```

@cheatfate concern was that by having to pass around the `RpcServer` object, we'd have to provide global `RpcServer` instances, which hurts more when implementing threading, etc. and also means it's more tying of modules together, as you have to import the module where the global `RpcServer` is defined, or else have all your RPCs in one module with the server definition.

Allowing the rpc's to be created without registering simplifies dependencies.

The original RPC macro is still there as `registerRpc`, and this covers the most simple case as you mentioned.
```nim
server.registerRpc("myRpc"):
  # ...
```

coffeepots:
Currently in order to support invoking `rpc` inside a proc all that's required is removing the `*` from the proc.

Doesn't it make sense to offer more choice with the API though? If people want to declare RPCs and manually attach them to server, there is no way to do this currently - you must pass an `RpcServer` object to take advantage of the automatic parameter marshalling.

coffeepots:
Closed in favour of not-exported the generated `rpc` procs, allowing creating RPCs in proc scope.

PR #13, opened 2018-06-04
Refactoring and conversion to use nim-asyncdispatch2.

cheatfate:


zah:
It's usually better to avoid mixing refactorings and module moves/renames in the same commit. This makes the pull request harder to review. The better approach is to move the code in a separate commit that doesn't introduce any other changes.

eth-rpc/serverdispatch.nim
```diff
@@ -0,0 +1,295 @@
...
+
+proc wrapReply*(id: JsonNode, value: JsonNode, error: JsonNode): JsonNode =
+  return %{"jsonrpc": %"2.0", "result": value, "error": error, "id": id}
+
+proc sendError*(client: StreamTransport, code: int, msg: string,
+                id: JsonNode, data: JsonNode = newJNull()) {.async.} =
+  ## Send error message to client
+  let error = %{"code": %(code), "message": %msg, "data": data}
+  ifDebug: echo "Send error json: ",
+                wrapReply(newJNull(), error, id).pretty & "\c\l"
```
zah: A note about performance:

Such string concatenations are inefficient with the current Nim, because they lead to copying all of the concatenated buffers. It's more efficient to replace the concatenation with extra arguments to `echo`:

``` nim
echo "foo", bar, "\c\l"
```

There is a disclaimer though: future version of Nim will be able to optimize the copy away by stealing the buffer of the temporary strings involved (similar to C++'s rvalue references).

eth-rpc/serverdispatch.nim
```diff
@@ -0,0 +1,295 @@
...
+    var rpcsrv = cast[RpcServer](udata)
+    var fut = processMessage(rpcsrv, client, line)
+    yield fut
+    if fut.failed:
+      let err = fut.readError
+      if err of RpcProcError:
+        let rpcerr = RpcProcError(err)
+        await client.sendError(rpcerr.code, rpcerr.msg, rpcerr.data)
+      else:
+        await client.sendError(SERVER_ERROR, "Error", %getCurrentExceptionMsg())
```
zah: Shouldn't this still extract the message from the `err` value instead of calling `getCurrentExceptionMsg`? I don't think calling `getCurrentExceptionMsg` outside of an `except` handler is a well-defined operation.

eth-rpc/clientdispatch.nim
```diff
@@ -1,43 +1,50 @@
-import asyncnet, asyncdispatch, tables, json, macros
```
zah: Since the client and server modules are part of the API, why don't we call them just `json_rpc/client` and `json_rpc/server`?

eth-rpc/serverdispatch.nim
```diff
@@ -0,0 +1,295 @@
...
+
+proc newRpcServer*(address: TransportAddress): RpcServer =
+  ## Create new instance of RPC server.
+  result = RpcServer(procs: newTable[string, RpcProc]())
+  result.server = createStreamServer(address, processClient, {ReuseAddr},
+                                     udata = cast[pointer](result))
```
yglukhov: It's unfortunate that such pretty high level api requires `cast[pointer]`. It makes it much harder to reason about safety.

eth-rpc/clientdispatch.nim
```diff
@@ -1,43 +1,50 @@
...
+proc call*(self: RpcClient, name: string,
+           params: JsonNode): Future[Response] {.async.} =
   ## Remotely calls the specified RPC method.
   let id = $self.nextId
   self.nextId.inc
-  let msg = $ %{"jsonrpc": %"2.0", "method": %name, "params": params, "id": %id} & "\c\l"
-  await self.socket.send(msg)
+  var msg = $ %{"jsonrpc": %"2.0", "method": %name,
+                "params": params, "id": %id} & "\c\l"
+  discard await self.transp.write(cast[pointer](addr msg[0]), len(msg))
```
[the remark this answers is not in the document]
yglukhov: +1. and `seq[byte]` probably?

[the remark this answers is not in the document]
coffeepots: Theses wrappers would need to be in `asyncdispatch2` ideally, rather than RPC.

eth-rpc/serverdispatch.nim
cheatfate: About what safety you talking about? Could you please be more precise and specific?

yglukhov: Talking about the memory safety in general. From this code it is obvious that a GC-ed object is passed as an untraced pointer to somewhere. So I need to take special care to inspect the lifetime of such untraced reference. Moreover, proper lifetimes have to be maintained in the future, which is just another thing to remember.

coffeepots: `cast` circumvents the type system (I presume this is what @yglukhov means), and is best kept in the internals of the source API.

In this case, the RPC API shouldn't need to know about how async is handling data, so the `cast` should be 'hidden' inside `asyndispatch` as an accessor such as a template such as (for example)

```nim
template write*(transport: StreamTransport, msg: string): untyped = transport.write(cast[pointer](addr msg[0]), len(msg))
```

This incurs no overhead, as the template is inlined, but keeps it so the RPC code doesn't need to know how `asyncdispatch` handles it's internal data.

coffeepots: So in this case (reasoning about the pointer), how would this be fixed? Surely it must use a pointer internally, and my suggestion of hiding this in fact makes reasoning more difficult since you don't know there's a cast.

Seems like the only way @yglukhov to avoid the reasoning issue is to not use pointers at all.

yglukhov: Will this be possible?
```nim
proc createStreamServer(adr, proc, flags, data: ref T): Server =
  GC_ref(data)
  createStreamServer(adr, proc, flags & {UnrefUdata}, data)
```
Then not forget to unref it where relevant.

eth-rpc/serverdispatch.nim
coffeepots: These will be replaced by `nim-chronicle` messages soon.

PR #16, opened 2018-06-07
Use asyncdispatch2 instead of asyncdispatch.

cheatfate:


eth-rpc/client.nim
```diff
@@ -72,32 +79,38 @@ proc connect*(self: RpcClient, address: string, port: Port): Future[void]
...
+  self.transp = await connect(self.address)
 
 proc connect*(self: RpcClient, address: string, port: Port) {.async.} =
-  await self.socket.connect(address, port)
-  self.address = address
-  self.port = port
+  # TODO: `address` hostname can be resolved to many IP addresses, we are using
+  # first one, but maybe it would be better to iterate over all IP addresses
+  # and try to establish connection until it will not be established.
+  let addresses = resolveTAddress(address & ":" & $int(port))
```
zah: Why don't we add more construction options for the Address type (i.e. taking an address and a port as separate parameters)? These unnecessary string concatenations keep popping up.

cheatfate: Yep, i will fix it, after fix Nim.

eth-rpc/server.nim
```diff
@@ -35,20 +36,15 @@ const
       (INVALID_REQUEST, "No id specified")
     ]
 
-template ifDebug*(actions: untyped): untyped =
-  # TODO: Replace with nim-chronicle
-  when not defined(release): actions else: discard
+when not defined(release):
```
coffeepots: I'm curious about this change. Is there a difference in code generation?

yglukhov: +1. Old variant looks cleaner to me

eth-rpc/server.nim
```diff
@@ -9,10 +12,8 @@ type
   # Procedure signature accepted as an RPC call by server
   RpcProc* = proc (params: JsonNode): Future[JsonNode]
 
-  RpcServer* = ref object
-    socket*: AsyncSocket
-    port*: Port
-    address*: string
+  RpcServer* = ref object of RootRef
```
yglukhov: Is there a reason for `RootRef`? By using `RootRef` `RpcServer` essentially becomes inheritable.

eth-rpc/client.nim
```diff
@@ -1,43 +1,46 @@
...
+           params: JsonNode): Future[Response] {.async.} =
   ## Remotely calls the specified RPC method.
   let id = $self.nextId
   self.nextId.inc
-  let msg = $ %{"jsonrpc": %"2.0", "method": %name, "params": params, "id": %id} & "\c\l"
-  await self.socket.send(msg)
+  let msg = $ %{"jsonrpc": %"2.0", "method": %name, "params": params,
+                "id": %id} & "\c\l"
+  let res = await self.transp.write(msg)
+  assert(res == len(msg))
```
yglukhov: This looks like a technical debt. Can't we just use `writeExact` or smth?

eth-rpc/server.nim
```diff
@@ -135,16 +136,73 @@ proc processClient(server: RpcServer, client: AsyncSocket) {.async.} =
         let err = future.readError[].ValueError
         await client.sendError(INVALID_PARAMS, err.msg, %"")
       else:
-        await client.sendError(SERVER_ERROR, "Error: Unknown error occurred", %"")
+        await client.sendError(SERVER_ERROR,
+                               "Error: Unknown error occurred", %"")
 
-proc serve*(server: RpcServer) {.async.} =
+proc newRpcServer*(address = "localhost", port: Port = Port(8545)): RpcServer =
```
yglukhov: Using a `seq[IpAddress]` type instead of string is better here. The only valid value for a string that is not an ip address is "localhost", which can easily be handled elsewhere. While `strings` interface limits the user in available configurations. E.g. the server can't be configured to listen on _all_ external interfaces, can it?

This interface could be left as a thin shortcut to a more elaborate interface, but I'm not really sure if that's needed.

eth-rpc/server.nim
cheatfate: With latest devel old variant gives me:
```
server.nim(86, 46) template/generic instantiation from here
server.nim(89, 3) template/generic instantiation from here
server.nim(48, 19) Error: obsolete usage of 'defined', use 'declared' instead
```

yglukhov:
As a side note, I would prefer longer lines. My text editor can soft wrap in case a line doesn't fit my monitor. ;) 13" monitor was pretty happy about the longer lines. 32" monitor looks empty now...

eth-rpc/server.nim
cheatfate: `address = "0.0.0.0"` all external interfaces via IPv4.
But please note this is was before this PR, so i just fixed behavior of `address: string`.

eth-rpc/client.nim
cheatfate: You never can know that you have send exactly requested amount of bytes, e.g. remote peer disconnects while we are sending message to it. `write()` procedure will of course try to send all available data, but as explained above, in some situations `write()` could not be completed successfully.

eth-rpc/server.nim
cheatfate: But what the reason not to make it inherited? Maybe some other developers wants to store more data in `RpcServer`, e.g. some state data.

eth-rpc/client.nim
yglukhov: Transport failure will fail the future and thus raise in the `await`. Is there any chance when `res != len(msg)`. If not, maybe `write` should return `Future[void]` instead? Or additionally be renamed to `writeExact`?

cheatfate:
We already considered to use `NEP-1`. Also 80 characters per line is a standard everywhere, it also nicely looked in github, because more longer lines introduce scrollbars.

eth-rpc/server.nim
yglukhov: Yes, you can't listen on all external interfaces on both ipv4 and ipv6. Since you're changing the interface anyway, why not change it to smth even better.

eth-rpc/server.nim
yglukhov: Ok, fair enough

eth-rpc/server.nim
yglukhov: This surely looks like some bug. Besides, I can't reproduce it with current devel (f99acdb07584d2c4c5eb1f22d998d97bcd823357).

eth-rpc/client.nim
cheatfate: `write()` returns Future[int] for a reason, so application can decide is it critical error, or not. When Future[void] is returned application could not know what happens and need to parse OSError to understand is it critical or not. Also there like 50% of chance not to send whole data, everything depends on network connection (remember we are doing this for mobiles, which can lost connection at any moment).

So this `assert` was made as `TODO:`, e.g. what RPC server must do if it not able to send whole data, maybe it needs to disconnect such client, or maybe it just needs to ignore it.

eth-rpc/server.nim
cheatfate: @yglukhov, please, I have not changed any interface of RPC.

yglukhov:
I never realized that this rule comes from NEP-1. Let it be subject for another discussion then.

eth-rpc/server.nim
yglukhov: Ah, right. Ok.

eth-rpc/server.nim
cheatfate: Updated my nim to 
```
Nim Compiler Version 0.18.1 [Windows: amd64]
Compiled at 2018-06-08
Copyright (c) 2006-2018 by Andreas Rumpf

git hash: f99acdb07584d2c4c5eb1f22d998d97bcd823357
active boot switches: -d:release
```
And when compiling via `nim c server.nim` i'm still getting error
```
server.nim(86, 46) template/generic instantiation from here
server.nim(89, 3) template/generic instantiation from here
server.nim(48, 19) Error: obsolete usage of 'defined', use 'declared' instead
```

eth-rpc/server.nim
yglukhov: Actually not that fair, because there's no way for the user to initialize his subclass. The only initializer we have is `newRpcServer: RpcServer`

eth-rpc/server.nim
yglukhov: Further investigation revealed a supposedly nim bug: https://github.com/nim-lang/Nim/issues/7997. Thus current solution may serve as a workaround for that, i guess.

eth-rpc/server.nim
cheatfate: Agree, this is one more TODO to have more optional init procedure.

eth-rpc/client.nim
zah: `asserts` are not to-do items in general, so we better place an explicit TODO comment on the line above.

eth-rpc/client.nim
```diff
@@ -1,43 +1,46 @@
...
+macro checkGet(node: JsonNode, fieldName: string,
+               jKind: static[JsonNodeKind]): untyped =
   let n = genSym(ident = "n") #`node`{`fieldName`}
   result = quote:
     let `n` = `node`{`fieldname`}
-    if `n`.isNil: raise newException(ValueError, "Message is missing required field \"" & `fieldName` & "\"")
-    if `n`.kind != `jKind`.JsonNodeKind: raise newException(ValueError, "Expected " & $(`jKind`.JsonNodeKind) & ", got " & $`node`[`fieldName`].kind)
+    if `n`.isNil:
+      raise newException(ValueError,
+                    "Message is missing required field \"" & `fieldName` & "\"")
```
zah: There was an attempt to improve the formatting here, but this new formatting looks quite unusual and strange. Is this the reason why Yuriy is asking about the NEP-1 80-char line limit? :P

eth-rpc/server.nim
zah: I guess @yglukhov is making a valid point that you might want to listen on multiple distinct interfaces (not just all of them). We can try to support such a `seq` parameter throughout the whole networking stack.

yglukhov: @zah, true, although as @cheatfate says this could be a subject for another pr.

coffeepots: These changes can be part of the refactoring to support network agnostic transports.

PR #19, opened 2018-06-26
Transport agnostic rpc

coffeepots:
This PR allows custom transport code to be used for RPC transmission.

In order to do this, we now have:
`defineRpcServerTransport(nameOfCallbackFunc)`
and
`defineRpcClientTransport(TransportType, AddressType, strPrefix)`

Both of these macros allow you to define the code for
  * `read`
  * `afterRead`
  * `write`
  * `close`

with the client also providing

  * `connect`

and the server providing

  * `error`, which allows altering data before being sent.

These are supplied in a verb-code format.
For example, here is the basics of an HTTP wrapper for the server:

```nim
defineRpcServerTransport(httpProcessClient):
  write:
    const contentType = "Content-Type: application/json-rpc"
    value = &"Host: {$transport.localAddress} {contentType} Content-Length: {$value.len} {value}"
    transport.write(value)
  afterRead:
    value = "HTTP Server".extractJsonStr(value)
```

Both constructs provide default code that essentially builds the previous RPC transports through StreamServer, supplied as `newRpcStreamServer` and `newRpcStreamClient`, which should work as a standard socket RPC server/client without further changes.

Additionally, data hex strings and quantity hex strings now each have their own distinct string type.

  * This enables data sanitising input/output from the RPC server.
  * Specifying HexDataStr or HexQuantityStr as input/output for an RPC proc will automatically get checked
    against the Ethereum standard and raise an exception if not valid.
  * Note that data and quantities have different checking requirements
    * For example: data must be an even length whereas quantities do not

TODO:
* HTTP server is currently just a mock up and needs further work, error & content checking, and so on.
* HTTPS support
* There is still some further work in testing other transports that don't rely on asyncdispatch sockets.


coffeepots:
Note: renamed from `eth-rpc` to `json_rpc`.

PR #21, opened 2018-07-06
Transport agnostic rpc no dsl

coffeepots:
This PR reworks the changes in #19 to simplify the transport agnostic RPC server (see last commit 
cf44cc5  for all changes compared to #19).

We now have:
* `router.nim`: Provides the existing `rpc` transformation which includes json marshalling, performs the method name look up and invokes the RPC itself, isolated from the server:
  * `proc ifRoute*(router: RpcRouter, data: JsonNode, fut: var Future[JsonNode]): bool`
  * `proc route*(router: RpcRouter, data: JsonNode): Future[JsonNode]`

* `server.nim`: Now just performs the async calls, json sanity checks, and holds common types.
  * `proc processMessages*[T](server: RpcServer[T], line: string): Future[string]` is the generic function to perform the server duties such as checking the json state and invoking the routing as above. This is the proc the transport specific code will call.

* `sockettransport.nim`: This is the module that provides the socket specific code.
  * `proc processClient(server: StreamServer, transport: StreamTransport)` is a bespoke proc to handle the socket transport details, where it is passed to `processMessages` in `server.nim` to continue the flow of data and invoke the RPC call handling.

* `client.nim` has been simplified to be socket only (no DSL to define it) but is otherwise unchanged.


tests/ethhexstrings.nim
```diff
@@ -0,0 +1,151 @@
...
+  # ignore the last character so we retain '0' on zero value
+  while cidx < value.len - 1 and value[cidx] == '0':
+    cidx.inc
+  value[cidx .. ^1]
+
+proc encodeQuantity*(value: SomeUnsignedInt): string =
+  var hValue = value.toHex.stripLeadingZeros
+  result = "0x" & hValue
+
+template hasHexHeader*(value: string | HexDataStr | HexQuantityStr): bool =
```
cheatfate: This is not a proper check. Please check this template code for `value == ""`.

json_rpc/sockettransport.nim
```diff
@@ -0,0 +1,142 @@
...
+      else:
+        await transport.sendError(SERVER_ERROR,
+                              "Error: Unknown error occurred", %"")
+    else:
+      let res = await future
+      result = transport.write(res)
+
+# Utility functions for setting up servers using stream transport addresses
+
+proc addStreamServer*(server: RpcServer[StreamServer], address: TransportAddress, callBack: StreamCallback) =
```
cheatfate: Why you need here `callBack` argument?

coffeepots: A relic of how it was previously done. The callback isn't needed to be provided here now, I'll update it.

tests/ethhexstrings.nim
coffeepots: Added :)

PR #25, opened 2018-07-14
HTTP-JSONRPC

cheatfate:
HttpRpcServer/HttpRpcClient and tests.


coffeepots:
Sorry @cheatfate I meant to include https://github.com/status-im/nim-http-utils but I see this is pulled through with `nimble test` anyway.

PR #31, opened 2018-08-15
Add ref int and ref int64

coffeepots:
Support for `ref int` and `ref int64`

### DIFF
--- a/json_rpc/jsonmarshal.nim
+++ b/json_rpc/jsonmarshal.nim
@@ -9,8 +9,11 @@ proc `%`*(n: byte{not lit}): JsonNode =
 proc `%`*(n: uint64{not lit}): JsonNode =
   result = newJInt(int(n))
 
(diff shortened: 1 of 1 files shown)


json_rpc/jsonmarshal.nim
```diff
@@ -9,8 +9,11 @@ proc `%`*(n: byte{not lit}): JsonNode =
 proc `%`*(n: uint64{not lit}): JsonNode =
   result = newJInt(int(n))
 
-proc `%`*(n: ref int|ref int64): JsonNode =
-  result = newJInt(int(n[]))
+proc `%`*(n: ref int | ref int64): JsonNode =
```
zah: Why not go all the way and say `ref SomeInteger`?

coffeepots: Good point!

PR #33, opened 2018-08-29
Turn some potentially dangerous templates into functions

zah:
Be careful when creating templates. If the input parameters are
referenced within the body multiple times, this may lead to multiple
evaluations of functions with side-effects.

### DIFF
--- a/json_rpc/jsonmarshal.nim
+++ b/json_rpc/jsonmarshal.nim
@@ -1,9 +1,9 @@
 import macros, json, options, typetraits
 
-template expect*(actual, expected: JsonNodeKind, argName: string) =
+proc expect*(actual, expected: JsonNodeKind, argName: string) =
   if actual != expected: raise newException(ValueError, "Parameter [" & argName & "] expected " & $expected & " but got " & $actual)
 
-template expectType*(actual: JsonNodeKind, expected: typedesc, argName: string, allowNull = false) =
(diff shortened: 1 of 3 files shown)


json_rpc/jsonmarshal.nim
```diff
@@ -1,9 +1,9 @@
 import macros, json, options, typetraits
 
-template expect*(actual, expected: JsonNodeKind, argName: string) =
+proc expect*(actual, expected: JsonNodeKind, argName: string) =
   if actual != expected: raise newException(ValueError, "Parameter [" & argName & "] expected " & $expected & " but got " & $actual)
 
-template expectType*(actual: JsonNodeKind, expected: typedesc, argName: string, allowNull = false) =
+proc expectType*(actual: JsonNodeKind, expected: typedesc, argName: string, allowNull = false) =
```
coffeepots: For this one, how does proc evaluate the `const` error message when different types are passed? I was under the impression `const` is only ever evaluated once for a `proc`?

zah: A `typedesc` param makes the proc generic. It will be compiled separately for each unique type being passed to it.

Perhaps there is a code bloat argument here against using a function (but the compiled bodies seem simple and specialized enough). An alternative fix would be to assign `actual` to a local variable inside the template.

yglukhov:
This looks still actual... @zah mind rebasing?

PR #35, opened 2018-11-07
add host to http header

jangko:
geth refuses to answer request if the http header doesn't contains 'host:'


cheatfate:
@jangko could you please try to change HTTP version from `1.1` to `1.0`, and test it without `Host` field.

jangko:
HTTP version 1.0 works without `Host` field

jangko:
@cheatfate should we stick to HTTP version 1.1 or change it to 1.0?

cheatfate:
HTTP/1.0 do not support `Host` but it also do not support `Connection: keep-alive`, so after every request server can/must drop connection from client, while HTTP/1.1 connection will not be dropped, so client can work faster on continuous requests.

PR #39, opened 2018-11-09
fixes #36, switch to HTTP/1.0 and drop `Connection: keep-alive` from httpclient

### LINKED ISSUE #36: httpclient doesn't support chunked transfer encoding when receiving response
for example, parity ethereum client uses `Transfer-Encoding: chunked` and no `Content-Length` field.
this make httpclient unusable to perform rpc

jangko:


coffeepots:
Also I'm wondering how easy it would be to add some simple tests for the chunking. @jangko would this be easy for you to set up?

jangko:
> would this be easy for you to set up?

honestly, I have no idea where to start, that's why I left it blank.
Does it require me to modify the server too?
any clues perhaps?

cheatfate:
@jangko chunked transfer encoding is not supported by `HTTP/1.0`, and because `geth` properly supports HTTP specification you can use `HTTP/1.0` requests and avoid `chunked` answers.

`nim-json-rpc` is also not the right place for `chunked` decoder.

jangko:
luckily, parity also support `HTTP/1.0`.
does it mean we also drop `Connection: keep-alive` field too when connecting with HTTP/1.0?
is that acceptable?

jangko:
switching to `HTTP/1.0` and removing `Connection: keep-alive` prove to be more problematic.
I think I will go through the hard way by putting it into httputils and implement chunked transfer decoding

jangko:
I didn't say about speed, I've located the problem inside asyncdispatch2, perhaps we can pending that chunked transfer problem later and for now we can switch to HTTP/1.0 and drop keep-alive connection.

coffeepots:
@jangko We've been discussing the chunking work and have come to the conclusion that we should stick with HTTP 1.0 for now (which you've already done here) and later have fully genericised chunking code further up in the tool chain.

With that in mind, can you remove the chunking code from this PR for now and will look at implementing it later.

coffeepots:
@jangko We've been discussing the chunking work and have come to the conclusion that we should stick with HTTP 1.0 for now (which you've already done here) and later have fully genericised chunking code further up in the tool chain.

jangko:
done

jangko:
closed in favor of 6a0b0ff30d602e59056d1558662667ae5700d09c

PR #41, opened 2018-11-12
fixes #37 add http method get/set to httpclient

### LINKED ISSUE #37: httpclient doesn't have feature to specify HTTP method to POST
for example, parity-ethereum requires HTTP POST method when using it's json-rpc facility

### LINKED ISSUE #38: httpserver need to implement CORS(Cross Origin Resource Sharing) policy
supposed the server is queried by a browser or by a debugging tool running inside a browser, enabling CORS will be handy.

link: [how to make cross domain request](https://www.moxio.com/blog/12/how-to-make-a-cross-domain-request-in-javascript-using-cors)

jangko:


cheatfate:
`HttpMethod` enum is present in imported `nim-http-utils`, there no need in declaration of new one enum.

zah:
Doesn't it make more sense to specify the HTTP method on each request (in other words, as an extra parameter to the `call` proc)? I guess the API description modules (e.g. `ethcallsig.nim` in the nimbus repo) could specify the HTTP method for each call with a [custom pragma](https://nim-lang.org/docs/manual.html#implementation-specific-pragmas-custom-annotations).

cheatfate:
The biggest problem is how this methods are handled by browsers. `GET` method answer is often get cached by browser while `POST` method answer is never cached by browser.

cheatfate:
This changes are needed for proper working with HTTP RPC client, but i do not like approach introduced in this PR. I wish not to have any additional methods in `RpcHttpClient` interface. Maybe generic `options` object would be more appropriate.

coffeepots:
Just wondering why the parameters added is `HTTPMethod` rather than `HttpClientOptions`, which includes the `HTTPMethod` state. Does it make more sense to have the additional parameters to `call` pass an `HttpClientOptions` object itself so that potentially in the future this can be expanded without requiring additional parameters (@cheatfate I assume this is what you're referring to as well)?

It just kinda seems like the `HttpClientOptions` is a bit 'wasted' in it's potential, in that it's just used to contain the `HTTPMethod` and isn't used anywhere else, yet it is exposed as a public type.

Having said that, I'm not sure on the usage pattern for passing GET/POST. @janko are you finding that all you need to change on `call` is the method type and creating a `HttpClientOptions` just for passing to `call` was overkill? Were you planning on expanding `HttpClientOptions` or did you create the type just to keep things separate inside the client?

jangko:
> Were you planning on expanding HttpClientOptions

yes

> I'm not sure on the usage pattern for passing GET/POST

Some server use GET only, some use POST, and the other support both, or they specify which rpc-method use GET while others rpc-method use POST.  

In the last scenario, we may need to call `call` proc using alternately between HttpMethod GET and POST.

> are you finding that all you need to change on call is the method type

I'm not sure

> and creating a `HttpClientOptions` just for passing to call was overkill?

no, but I think it will be inconvenient

coffeepots:
Agree it is a bit inconvenient to pass things like `HttpClientOptions(httpMethod: MethodGet)` (or worse having to create the object on another line just to pass it), and the API can always be expanded later if necessary :)

zah:
Actually, I suggested to @jangko that POST is considered the default choice/requirement for all RPC calls, due to some reasoning provided here:
https://www.simple-is-better.org/json-rpc/transport_http.html

Using POST on all calls by default will be compatible with both Geth and Parity, so another follow-up commit may be added to simplify the API.

coffeepots:
As always, things are never simple, right? :)

It sounds like POST is a sensible default but favouring control seems like the best option.

PR #42, opened 2018-11-12
add optional arg support to rpc macro

### LINKED ISSUE #40: rpc macro needs to support optional parameters
for example:

```Nim
rpcsrv.rpc("debug_traceTransaction") do(data: HexDataStr, options: Option[TraceTxOptions]) -> int:
  discard
```
current implementation will throw error: `errorMessage="Expected 2 Json parameter(s) but got 1"`

expected behavior: it is ok to give only one parameter

jangko:
fixes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Here is a more complex parameter example:
 ```nim
 type
   HeaderKind = enum hkOne, hkTwo, hkThree
-  
+
   Header = ref object
     kind: HeaderKind
     size: int64
@@ -88,7 +88,7 @@ type
     name: string
 
 router.rpc("updateData") do(myObj: MyObject, newData: DataBlob) -> DataBlob:
-  result = myObj.data  
+  result = myObj.data
   myObj.data = newData
 ```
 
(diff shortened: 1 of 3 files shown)


coffeepots:
I do have a lingering concern though that we might want an `Option[T]` type in the middle of a parameter list.

A possible future change that might allow more flexibility for the user, is removing the error when there are non-option types after an `Option[T]`, and instead simply only take `Option[T]` parameters at the end as optional.

For example with this: `p1: int, p2: Option[bool], p3: int, p3: Option[int], p4: Option[string]`, to allow `p3` and `p4` to be optional but `p3` must be specified as `none`.

jangko:
it also can be further enhanced to accept not only parameters list(array), but accept parameter dictionary(json object) with named parameters instead of positional parameters. In that way, we can offer greater flexibility to user. Actually JSON RPC 2.0 spec allow this:

> 4.2 Parameter Structures
> If present, parameters for the rpc call MUST be provided as a Structured value. Either by-position through an Array or by-name through an Object.
> *    by-position: params MUST be an Array, containing the values in the Server expected order.
> *   by-name: params MUST be an Object, with member names that match the Server expected parameter names. The absence of expected names MAY result in an error being generated. The names MUST match exactly, including case, to the method's expected parameters.

but perhaps it is not as urgent as optional parameter

coffeepots:
Technically we can do parameter by name already without changing anything, if you have a file of RPC signatures you pass to `createRpcSigs`.

Once this converts them into client RPC calls you can use `client.myRpc(param1 = 1, param5 = 5)`.

PR #43, opened 2018-11-23
Options fixes

coffeepots:
* Fix for parsing arguments so it doesn't read the return type (shows as `index out of bounds` when just one optional argument)
* Removed `expectType` as it only handled built in types for `Option[T]`. Instead, the normal `fromJson` machinery is used and checks for type. This allows things like `Option[MyObject]` which wasn't possible before.
* Added tests for non-builtin types in `Option[T]`, checking the single `Option[T]` parameter, and embedding `Option[T]` for a couple of layers.


coffeepots:
@Zah I've just added a fix for @jangko's comment above, so now manually constructed object fields of type `Option[T]` are handled if the user hasn't provided a `JNull` for the field.

PR #47, opened 2019-01-10
add "--threads:on" to tests

stefantalpalaru:


zah:
The Windows compilation failure needs to be resoled.

stefantalpalaru:
It's actually an asyncdispatch2 problem, so it will be solved there. We'll trigger another CI run here afterwards to make sure.

stefantalpalaru:
Ready for review, @zah.

PR #54, opened 2019-05-09
Check if method field is a string against unhandled exception

kdeme:
`FieldError` exception can occur [here](https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/router.nim#L113) when a non string value is send for the json method member.

Would result in a crash for `socketserver`. `httpserver` would be fine due to its catch all code.


stefantalpalaru:
The AppVeyor build seems stuck with this PR: https://ci.appveyor.com/project/nimbus/nim-json-rpc/builds/24424478/job/mfewkdn2ftfss06t

stefantalpalaru:
Never mind, it always gets stuck in the same place: https://ci.appveyor.com/project/nimbus/nim-json-rpc/builds/24184247/job/fim1chtwr23pl264

PR #57, opened 2019-06-12
Websocket client support and some fixes

yglukhov:


json_rpc/jsonmarshal.nim
```diff
@@ -62,8 +69,11 @@ proc fromJson(n: JsonNode, argName: string, result: var bool) =
   result = n.getBool()
 
 proc fromJson(n: JsonNode, argName: string, result: var int) =
-  n.kind.expect(JInt, argName)
-  result = n.getInt()
+  if n.kind == JString:
```
kdeme: Is there a reason why you need to do this conversion here? I think web3.utils generally is meant for this work.

While I don't see an immediate risk in this case, I believe it would be beneficial if we keep these `fromJson` calls as simple as possible, and that would probably mean accepting only 1 type in general.

json_rpc/jsonmarshal.nim
```diff
@@ -110,6 +120,14 @@ proc fromJson[T](n: JsonNode, argName: string, result: var seq[T]) =
     fromJson(n[i], argName, result[i])
 
 proc fromJson[N, T](n: JsonNode, argName: string, result: var array[N, T]) =
+  when T is byte:
```
kdeme: same remark as above is valid here

json_rpc/clients/httpclient.nim
```diff
@@ -176,60 +171,36 @@ proc httpMethod*(client: RpcHttpClient): HttpMethod =
 proc httpMethod*(client: RpcHttpClient, m: HttpMethod) =
   client.options.httpMethod = m
 
-proc call*(client: RpcHttpClient, name: string,
-           params: JsonNode, httpMethod: HttpMethod): Future[Response] {.async.} =
+method call*(client: RpcHttpClient, name: string,
```
kdeme: Do I understand right that the client was originally reconnecting automatically each time in while loop after the request ended and that now this gets done properly on the request call itself?

json_rpc/jsonmarshal.nim
yglukhov: Well in this particular case we could just change the method signatures from using `int` to `string`, not a big deal.

json_rpc/jsonmarshal.nim
yglukhov: This one however is not as simple. Currently there are a lot of values (like hashes and addresses) which are passed around as json strings, but in nim they are essentially byte arrays. One way to mitigate it would be to change their nim representations to `stint.uint`s, but I find it suboptimal, because (a) semantically they are not numbers and (b) conversion from hex string to uints and back is a bit more involved than for byte arrays, and is not really needed.

I think the most proper way would be to use distinct types for that and to define `fromJaon` overloads for them, but it felt like too much of a change at this point.

json_rpc/clients/httpclient.nim
yglukhov: "the client was originally reconnecting automatically each time in while loop" - that was the intended behavior I think, but unfortunately it did not work. Sending multiple requests in a row was subjected to a race condition when the send proc tried to write to the socket, (temporarily) closed by the read loop. Since we're not using HTTP 1.1 persistent connection anyway I thought this simplification was a good idea. Funnily I did it before I found that HTTP proto is now deprecated altogether in web3, but oh well.

json_rpc/jsonmarshal.nim
kdeme: That is how it is done in nimbus rpc by the way: https://github.com/status-im/nimbus/blob/master/nimbus/rpc/hexstrings.nim

kdeme: It is possible here to trigger an assertion from byteutils.  More specifically this one: https://github.com/status-im/nim-byteutils/blob/master/byteutils.nim#L51
Just provide as value e.g. `0x0000000000`. Basically it should be checked for `len*2` instead of `len` I believe.

Next to that, other wrong values such as "0x00" will not return with a specific "hex string error" but also still do the array check. This gives a confusing error message.

There is also different behaviour between the array input and string input version, in that the array version excepts it when N is not reached. (could argue that is faulty behaviour for the array version)

Anyway, this can be fixed, I know. However, the point I'm trying to make is that these, at first sight small additions, can create complexities that are not needed for all the "users" of nim-json-rpc.

json_rpc/jsonmarshal.nim
kdeme: I had a further look at `parseHexInt`, turns out it does not check for overflows. 
I didn't check in detail but I guess values such as `0x8000000000000000` and bigger will wrap around, not the behaviour you would want. (or at least different behaviour from providing currently an int that is too large)

json_rpc/clients/httpclient.nim
```diff
@@ -176,60 +171,36 @@ proc httpMethod*(client: RpcHttpClient): HttpMethod =
...
+  let transp = await connect(client.addresses[0])
+  var reqBody = $rpcCallNode(name, params, id)
+  echo "Sending (", client.httpMethod, "): ", reqBody
+  let res = await transp.sendRequest(reqBody, client.httpMethod)
   if not res:
     debug "Failed to send message to RPC server",
-          address = client.transp.remoteAddress(), msg_len = len(value)
-    await client.transp.closeWait()
+          address = transp.remoteAddress(), msg_len = len(reqBody)
+    transp.close()
```
cheatfate: Please, do not change `await closeWait()` to `close()`.

json_rpc/clients/socketclient.nim
```diff
@@ -49,3 +49,7 @@ proc connect*(client: RpcSocketClient, address: string, port: Port) {.async.} =
   client.transport = await connect(addresses[0])
   client.address = addresses[0]
   client.loop = processData(client)
+
+method close*(client: RpcSocketClient) =
+  # TODO: Stop the processData loop
+  client.transport.close()
```
cheatfate: please use `await closeWait()` here.

json_rpc/clients/websocketclient.nim
```diff
@@ -0,0 +1,54 @@
...
+  self.awaiting[id] = newFut
+  result = await newFut
+
+proc processData(client: RpcWebSocketClient) {.async.} =
+  while true:
+    while true:
+      var value = await client.transport.receivePacket()
+      if value == "":
+        # transmission ends
+        client.transport.close()
```
cheatfate: Please use `await closeWait()` here.

json_rpc/clients/websocketclient.nim
```diff
@@ -0,0 +1,54 @@
...
+    client.transport = await newWebSocket(client.uri)
+
+proc connect*(client: RpcWebSocketClient, uri: string) {.async.} =
+  client.transport = await newWebSocket(uri)
+  client.uri = uri
+  client.loop = processData(client)
+
+method close*(client: RpcWebSocketClient) =
+  # TODO: Stop the processData loop
+  client.transport.close()
```
cheatfate: Please use `await closeWait()` here.

json_rpc/clients/httpclient.nim
```diff
@@ -176,60 +171,36 @@ proc httpMethod*(client: RpcHttpClient): HttpMethod =
...
   else:
-    debug "Message sent to RPC server", address = client.transp.remoteAddress(),
-          msg_len = len(value)
-    trace "Message", msg = value
+    debug "Message sent to RPC server", address = transp.remoteAddress(),
+          msg_len = len(reqBody)
+    trace "Message", msg = reqBody
+
+  var value = await transp.recvData()
+  transp.close()
```
cheatfate: Please use `await closeWait()` here.

json_rpc/clients/websocketclient.nim
cheatfate: Please use `await closeWait()` here.

PR #61, opened 2019-07-02
Allow custom headers in ws backend. Set Origin to http://localhost by default

yglukhov:


yglukhov:
@kdeme yes, I think so.

PR #64, opened 2019-07-17
Lock ws provider to allow using it from different coroutines

yglukhov:


yglukhov:
Sorry, I changed my mind. Would be better and more efficient to take care of it in the `news` package.

PR #65, opened 2019-07-19
fix some problems raised by Nim 0.20.2

stefantalpalaru:


json_rpc/client.nim
```diff
@@ -82,9 +82,10 @@ proc processMessage*(self: RpcClient, line: string) =
   elif "method" in node:
     # This could be subscription notification
     let name = node["method"].getStr()
-    let handler = self.methodHandlers.getOrDefault(name)
-    if not handler.isNil:
-      handler(node{"params"})
+    {.gcsafe.}:
```
zah: Let's add the practice that every `gcsafe` override must come with a comment explaining the reason behind it.
I guess the problem here is that the proc signature type of `handler` is not marked as `gcsafe`.

stefantalpalaru: You were right. Fixed.

PR #69, opened 2019-11-22
Better error handling

yglukhov:
- Fail pending futures on error in read loop
- Added `onDisconnect` handler.

### DIFF
--- a/json_rpc/clients/websocketclient.nim
+++ b/json_rpc/clients/websocketclient.nim
@@ -1,4 +1,4 @@
-import ../client, chronos, tables, json, strtabs, chronicles
+import ../client, chronos, tables, json, strtabs
 
 const newsUseChronos = true
 include news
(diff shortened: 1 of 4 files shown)


json_rpc/clients/websocketclient.nim
```diff
@@ -56,11 +68,7 @@ proc connect*(client: RpcWebSocketClient, uri: string, headers: StringTableRef =
...
-  client.loop.addCallback do(data: pointer):
-    if client.loop.failed:
-      let err = client.loop.readError()
-      error "websocket rpc", msg = err.msg, stacktrace = err.getStackTrace()
 
 method close*(client: RpcWebSocketClient) {.async.} =
-  # TODO: Stop the processData loop
-  client.transport.close()
+  if not client.transport.isNil:
+    client.loop.cancel()
```
kdeme: I don't fully understand this. Doesn't this need a cancellation callback implementation that closes the transport? Or is there one in news? Or where else is this happening?

yglukhov: The transport is closed in `processData` when it breaks out of the read loop either when we cancel it or it reads EOF.

kdeme: Ah yes, via the CancelledError. Thanks for the explainer!

Looking into it did make me notice that `CancelledError` is not a `CatchableError` currently in Chronos: https://github.com/status-im/nim-chronos/blob/master/chronos/asyncfutures2.nim#L56 
So we have a problem there...

yglukhov: Good catch! Chronos PR: https://github.com/status-im/nim-chronos/pull/59

PR #72, opened 2020-03-17
Refactor the router to allow bypassing the use of std/json

zah:
The users will be now able to write RPC handlers returning the
distinct type `StringOfJson`. This will bypass any use of the
standard library's JsonNode type and its serialization routines.

The change was motivated by the integration of JSON-RPC in
status-im/nim-beacon-chain where most of the data types cannot
be handled by Nim's std lib or json-rpc's jsonmarshal module.


json_rpc/router.nim
```diff
@@ -226,35 +219,28 @@ macro rpc*(server: RpcRouter, path: string, body: untyped): untyped =
...
+    proc `doMain`(`paramsIdent`: JsonNode): Future[`ReturnType`] {.async.} =
+      `setup`
+      `procBody`
 
-    # delegate async proc allows return and setting of result as native type
+  if ReturnType == ident"JsonNode":
+    # `JsonNode` results don't need conversion
+    result.add quote do:
+      proc `procName`(`paramsIdent`: JsonNode): Future[StringOfJson] {.async, gcsafe.} =
+        return StringOfJson($(await `doMain`(`paramsIdent`)))
```
kdeme: Does this mean that we are no longer trapping the exceptions?

zah: No, I moved the exception handling in the code that is dispatching the request to the right user-supplied proc:
https://github.com/status-im/nim-json-rpc/pull/72/files#diff-380a803ec724e6ae58db2c9dcf966219R127-R135

This makes the overall code smaller and leads to less cyclomatic complexity in the invisible generated code (something I consider good for any macro-generated code in general).

PR #78, opened 2020-05-29
feature/websocket-server

### LINKED ISSUE #76: Add websocket server support
There is already websocket client support through the news package: https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/clients/websocketclient.nim

Could do similar for the server side with [news](https://github.com/Tormund/news).
But in the examples there `asynchttpserver` is used. So, we would have to do something similar but adjust it so that is uses the the httpserver of nim-json-rpc.

decanus:
closes #76

### DIFF
--- a/json_rpc/rpcserver.nim
+++ b/json_rpc/rpcserver.nim
@@ -1,3 +1,3 @@
 import server
-import servers/[socketserver, httpserver]
-export server, socketserver, httpserver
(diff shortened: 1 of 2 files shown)


arnetheduck:
obsolete - a new chronos-based websocket server is being developed which will be used here

PR #82, opened 2020-08-06
can't use `result` when returning Future[void]

stefantalpalaru:


stefantalpalaru:
See also: https://github.com/status-im/nim-chronos/issues/118

stefantalpalaru:
CI will be broken here, until this is merged: https://github.com/Tormund/news/pull/18

PR #86, opened 2020-10-07
added a check for the content length of responses - fixes #1650 of nbc

onqtam:
https://github.com/status-im/nim-beacon-chain/issues/1650

### DIFF
--- a/json_rpc/clients/httpclient.nim
+++ b/json_rpc/clients/httpclient.nim
@@ -16,7 +16,7 @@ type
 
 const
   MaxHttpHeadersSize = 8192        # maximum size of HTTP headers in octets
-  MaxHttpRequestSize = 128 * 1024  # maximum size of HTTP body in octets
+  MaxHttpResponseSize = 128 * 1024  # maximum size of HTTP body in octets
   HttpHeadersTimeout = 120.seconds # timeout for receiving headers (120 sec)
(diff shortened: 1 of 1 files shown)


json_rpc/clients/httpclient.nim
```diff
@@ -16,7 +16,7 @@ type
 
 const
   MaxHttpHeadersSize = 8192        # maximum size of HTTP headers in octets
-  MaxHttpRequestSize = 128 * 1024  # maximum size of HTTP body in octets
+  MaxHttpResponseSize = 128 * 1024  # maximum size of HTTP body in octets
```
arnetheduck, line 19: where is this limit coming from? it looks rather arbitrary - ie requesting a block or state over RPC will easily exceed this limit

onqtam: This PR was a direct response to https://github.com/status-im/nimbus-eth2/issues/1650 
Perhaps it should be way bigger by default and should also be configurable (as opposed to being a simple const)

arnetheduck: Indeed - it is not a practical solution to use a constant here - the constant could be a default, but there must be an override at least per url or it won't work for several actual requests we issue (and be unnecessarily lax for others)

arnetheduck:
replaced by #95

PR #88, opened 2020-11-12
Added marshalling support for uint32

jm-clius:
Added support for uint32 marshalling as required for [Waku v2 API](https://github.com/status-im/nim-waku/issues/258).
See also [this pull request](https://github.com/status-im/nim-waku/pull/263#pullrequestreview-528496942) and discussion with @kdeme 
Thanks! :)

### DIFF
--- a/json_rpc/jsonmarshal.nim
+++ b/json_rpc/jsonmarshal.nim
@@ -34,6 +34,7 @@ proc fromJson*[T](n: JsonNode, argName: string, result: var seq[T])
 proc fromJson*[N, T](n: JsonNode, argName: string, result: var array[N, T])
 proc fromJson*(n: JsonNode, argName: string, result: var int64)
 proc fromJson*(n: JsonNode, argName: string, result: var uint64)
+proc fromJson*(n: JsonNode, argName: string, result: var uint32)
 proc fromJson*(n: JsonNode, argName: string, result: var ref int64)
 proc fromJson*(n: JsonNode, argName: string, result: var ref int)
 proc fromJson*[T](n: JsonNode, argName: string, result: var Option[T])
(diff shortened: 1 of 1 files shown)


json_rpc/jsonmarshal.nim
```diff
@@ -94,6 +95,16 @@ proc fromJson*(n: JsonNode, argName: string, result: var uint64) =
       ValueError, "JSON-RPC input is an unexpected negative value")
   result = uint64(asInt)
 
+proc fromJson*(n: JsonNode, argName: string, result: var uint32) =
+  n.kind.expect(JInt, argName)
+  let asInt = n.getInt()
```
arnetheduck, line 100:
```suggestion
  let asInt = n.getBiggestInt()
```

to avoid 31-bit issues on 32-bit platforms

json_rpc/jsonmarshal.nim
```diff
@@ -94,6 +95,16 @@ proc fromJson*(n: JsonNode, argName: string, result: var uint64) =
...
+  # signed -> unsigned conversions are unchecked
+  # https://github.com/nim-lang/RFCs/issues/175
+  if asInt < 0:
+    raise newException(
+      ValueError, "JSON-RPC input is an unexpected negative value")
+  result = uint32(asInt)
```
arnetheduck, line 106: we try to avoid using `result`, it disables several useful compiler diagnostics

jm-clius: Makes sense. I'd suggest changing it to `argValue`, but I'm hesitant to do so as it will deviate from the pattern of all the other `fromJson*` in the same file and the[ documentation here](https://github.com/status-im/nim-json-rpc#parameters). Otherwise I could change it for everything else too?

jm-clius:
Hey @arnetheduck, not sure if you saw my comment re using `result` - I'm hesitant to deviate from the docs and the pattern established in the rest of the module, especially if it complicates future refactoring. Let me know what you think :)

json_rpc/jsonmarshal.nim
arnetheduck: right,  I missed that this was a `var` parameter - my comment mainly applies to the magic implicit variable - though reusing the name clearly causes this kind of confusion - it's really an unfortunate feature of nim's - but we can leave it for a separate PR to refactor at some point - there's a lot of legacy code out there that follows practises that have been shown to not work well

jm-clius: Ah, great. Thanks, Jacek! Yeah, I've discovered a number of these Nim intricacies, though in general I'm quite enjoying its flexibility.
Can this PR be merged then, or should we log the refactor work first?

arnetheduck (approved):

PR #95, opened 2021-02-11
fixes

arnetheduck:
* fix https://github.com/status-im/nimbus-eth2/issues/1650
* only one of `result` and `error` allowed in response
* fix invalid `string` instances being created from byte sequences
* fix large int64 parsing on 32-bit
* fix exception inheritance
* fix some dangling results
* some cleanups


json_rpc/clients/httpclient.nim
```diff
@@ -34,12 +37,10 @@ proc sendRequest(transp: StreamTransport,
...
   try:
-    let res = await transp.write(cast[seq[char]](request))
-    if res != len(request):
-      result = false
-    result = true
-  except:
-    result = false
+    let res = await transp.write(request.toBytes())
+    return res == len(request):
+  except CatchableError:
```
cheatfate, line 42: If you want to use `CatchableError` handler only you, you will catch `CancelledError` too.
Add `except CancelledError: raise exc` too.

cheatfate:
Also i will recommend to replace this http server implementation with http server from chronos. Because this one supports limited amount of features.

json_rpc/clients/httpclient.nim
arnetheduck: fixed, though the exception issues in the code run much deeper than this

arnetheduck:
> Also i will recommend to replace this http server implementation with http server from chronos. Because this one supports limited amount of features.

while that's reasonable, it's also beyond the scope of this PR - ie the focus here is to avoid the worst offenders in terms of json-rpc spec compatibility

tests/testerrors.nim
```diff
@@ -1,80 +0,0 @@
-#[
-  This module uses debug versions of the rpc components that
```
zah, line 2: How come we don't consider the removal of these tests a regression?

json_rpc/client.nim
```diff
@@ -34,62 +31,38 @@ method call*(client: RpcClient, name: string,
...
 
   if "id" in node:
-    let id = checkGet(node, "id", JInt)
+    let id = node{"id"} or newJNull()
 
     var requestFut: Future[Response]
-    if not self.awaiting.pop(id, requestFut):
-      raise newException(ValueError,
-        "Cannot find message id \"" & $node["id"].getInt & "\"")
+    if not self.awaiting.pop(id.getInt(-1), requestFut):
```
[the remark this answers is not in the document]
arnetheduck: this is the `id` in the response that we set ourselves in the request - if it's not an int, the server is messing with us

json_rpc/client.nim
```diff
@@ -34,62 +31,38 @@ method call*(client: RpcClient, name: string,
...
-
-    let errorNode = node{"error"}
-    if errorNode.isNil or errorNode.kind == JNull:
-      var res = node{"result"}
-      if not res.isNil:
-        requestFut.complete((false, res))
-      else:
-        requestFut.fail(newException(InvalidResponse, "Missing `result` field"))
+      requestFut.fail(newException(ValueError,
+        "Unsupported version of JSON, expected 2.0, received \"" & version & "\""))
```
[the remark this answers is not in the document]
arnetheduck: it would take a bunch of code because stdlib `json.$` crashes when given a `nil` node which `{}` returns, didn't seem like it was worth the effort for the relatively fringe problem here

tests/testerrors.nim
arnetheduck: they were never run (not part of `tests/all.nim`) - they also didn't match the behavior of the existing code - neither before nor after these fixes - in particular, they described that the `error` field in `Response` was used, when in fact an exception was being raised.

json_rpc/client.nim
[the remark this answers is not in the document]
arnetheduck: see also https://github.com/status-im/nim-json-rpc/blob/4eb39203ebd391c77d16a1c387dc8a6b7d90bc69/json_rpc/router.nim#L65 (which incidentally was also wrong - we were setting `id` inside the error object which is not how it looks in the spec)

PR #96, opened 2021-03-02
encode large uint64 as string

arnetheduck:
* allow json strings as integers
* put exception error message in error


arnetheduck:
closing as this brings backwards compat issues - also, upstream has (broken) uint64 support which interferes with what goes on here

PR #104, opened 2021-06-23
replace news with nim-ws

jangko:
also implement json rpc websocket server.

### DIFF
--- a/json_rpc/clients/websocketclient.nim
+++ b/json_rpc/clients/websocketclient.nim
@@ -2,6 +2,7 @@ import
   std/[strtabs, tables, uri, strutils],
   pkg/[chronos, ws/ws, chronicles],
   ws/extensions/compression/deflate,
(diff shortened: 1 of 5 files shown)


json_rpc/clients/websocketclient.nim
```diff
@@ -42,18 +42,20 @@ method call*(self: RpcWebSocketClient, name: string,
...
-      if value == "":
+    while ws.readystate != ReadyState.Closed:
+      var value = await ws.recv()
+
+      if value.len == 0:
         # transmission ends
         break
 
-      client.processMessage(value)
+      client.processMessage(cast[string](value))
```
arnetheduck, line 54: use `string.fromBytes` from https://github.com/status-im/nim-stew/blob/master/stew/byteutils.nim for string conversions

jangko: fixed

PR #101, opened 2021-05-29
Switch to the Chronos HTTP client (adds support for HTTPS)

zah:

PR #105, opened 2021-06-24
[FEATURE] Add http json rpc proxy

KonradStaniec:
Add possibility to create proxies by adding recovery proc to router.

Advantages of this approach:
- no need for serialization/deserialization of arguments and results
- can easy be adapted to different transports
- can be used for other means

Disadvantages:
- obscures a control flow a little bit
- not easy to decide about error control flow i.e should router handle errors thrown by recovery procdure (I went with this), or procedure should handle errors it self

Nim related questions:
- Is passing procedures around considered good nim style ?
- Is using `Optional` types considered good style ? 
- Can `Future` throw exceptions ? In `scala` world `Futures` does not throw exceptions and in its finished state they can be in either `success` state or `failed` state.

Projects related questions:
- From what i see we do not have `https` client yet ( correct me if I am wrong). This is kinda problematic, as infura currently forces `https` connections on all requests, so If we would like to use proxy to call infura we would need to implements https support. (or use websockets if we have such option)

### DIFF
--- a/json_rpc/router.nim
+++ b/json_rpc/router.nim
@@ -12,11 +12,11 @@ type
   # Procedure signature accepted as an RPC call by server
   RpcProc* = proc(input: JsonNode): Future[StringOfJson] {.gcsafe, raises: [Defect, CatchableError].}
 
-  RecoveryProc* = proc(name: string, input: JsonNode): Future[JsonNode] {.gcsafe, raises: [Defect, CatchableError].}
+  ProxyCall* = proc(name: string, input: JsonNode): Future[JsonNode] {.gcsafe, raises: [Defect, CatchableError].}
 
   RpcRouter* = object
     procs*: Table[string, RpcProc]
-    recoveryCall*: Option[RecoveryProc]
+    proxyCalls*: Table[string, ProxyCall]
 
 const
   methodField = "method"
@@ -33,17 +33,22 @@ const
 
 proc init*(T: type RpcRouter): T = discard
 
-proc init*(T: type RpcRouter, recoveryCall: RecoveryProc): T = RpcRouter(recoveryCall: some(recoveryCall))
-
 proc newRpcRouter*: RpcRouter {.deprecated.} =
   RpcRouter.init()
 
 proc register*(router: var RpcRouter, path: string, call: RpcProc) =
   router.procs.add(path, call)
 
-proc clear*(router: var RpcRouter) = router.procs.clear
+proc registerProxy*(router: var RpcRouter, path: string, call: ProxyCall) =
+  router.proxyCalls.add(path, call)
+
+proc clear*(router: var RpcRouter) = 
+  router.procs.clear
+  router.proxyCalls.clear
 
 proc hasMethod*(router: RpcRouter, methodName: string): bool = router.procs.hasKey(methodName)
+
+proc hasProxy*(router: RpcRouter, methodName: string): bool = router.proxyCalls.hasKey(methodName)
 
 func isEmpty(node: JsonNode): bool = node.isNil or node.kind == JNull
 
@@ -76,15 +81,14 @@ proc route*(router: RpcRouter, node: JsonNode): Future[StringOfJson] {.async, gc
     return wrapError(INVALID_REQUEST, "'method' missing or invalid")
 
   let rpcProc = router.procs.getOrDefault(methodName)
-
+  let proxyCall = router.proxyCalls.getOrDefault(methodName)
   let params = node.getOrDefault("params")
 
-  if rpcProc == nil and (not router.recoveryCall.isSome()):
+  if rpcProc == nil and (proxyCall == nil):
     return wrapError(METHOD_NOT_FOUND, "'" & methodName & "' is not a registered RPC method", id)
-  elif rpcProc == nil and (router.recoveryCall.isSome()):
+  elif rpcProc == nil and (proxyCall != nil):
     try:
-      let recoveryCall = router.recoveryCall.unsafeGet
-      let res = await recoveryCall(methodName, if params == nil: newJArray() else: params)
+      let res = await proxyCall(methodName, if params == nil: newJArray() else: params)
       let asString = StringOfJson($res)
       return wrapReply(id, asString)
 
--- a/json_rpc/router.nim
+++ b/json_rpc/router.nim
@@ -12,8 +12,11 @@ type
   # Procedure signature accepted as an RPC call by server
   RpcProc* = proc(input: JsonNode): Future[StringOfJson] {.gcsafe, raises: [Defect, CatchableError].}
 
+  RecoveryProc* = proc(name: string, input: JsonNode): Future[JsonNode] {.gcsafe, raises: [Defect, CatchableError].}
+
   RpcRouter* = object
     procs*: Table[string, RpcProc]
+    recoveryCall*: Option[RecoveryProc]
 
 const
   methodField = "method"
@@ -29,6 +32,8 @@ const
   defaultMaxRequestLength* = 1024 * 128
 
 proc init*(T: type RpcRouter): T = discard
+
+proc init*(T: type RpcRouter, recoveryCall: RecoveryProc): T = RpcRouter(recoveryCall: some(recoveryCall))
 
 proc newRpcRouter*: RpcRouter {.deprecated.} =
   RpcRouter.init()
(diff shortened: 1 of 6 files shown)


arnetheduck:
> Is passing procedures around considered good nim style ?

it's fine, most of the time, but it introduces a runtime choice that we sometimes prefer to avoid - some information here: https://status-im.github.io/nim-style-guide/language.proctypes.html

> Is using Optional types considered good style ?

Yes, as is using `Result`

> Can Future throw exceptions

`Future` itself can panic if it hasn't completed and you try to get the value - it can also raise exceptions when being read - under the hood, every time you call `await` it's a possibility an exception will be raised because the value is read implicitly there.

> in its finished state they can be in either success state or failed state.

Futures in chronos can also be cancelled.

> From what i see we do not have https client yet ( correct me if I am wrong). 

https://github.com/status-im/nim-json-rpc/pull/101

json_rpc/rpcproxy.nim
```diff
@@ -0,0 +1,38 @@
+import servers/[socketserver, httpserver]
```
kdeme, line 1: We typically write `import` once and then newline, and differentiating stdlib imports, module imports and local imports with again new lines. Local imports should use relative paths also.

See also: https://status-im.github.io/nim-style-guide/language.import.html

json_rpc/rpcproxy.nim
```diff
@@ -0,0 +1,38 @@
...
+import clients/[socketclient, httpclient]
+
+type RpcHttpProxy* = ref object of RootRef
+    rpcHttpClient: RpcHttpClient
```
kdeme, line 5: All your indentations seem off, 4 space vs 2?

json_rpc/rpcproxy.nim
```diff
@@ -0,0 +1,38 @@
...
+    rpcHttpServer: RpcHttpServer
+
+proc proxyCall(client: RpcHttpClient): ProxyCall =
+  result =
```
kdeme, line 9: For new code we avoid using result: https://status-im.github.io/nim-style-guide/language.result.html

json_rpc/rpcproxy.nim
```diff
@@ -0,0 +1,38 @@
...
+
+proc new*(T: type RpcHttpProxy, listenAddresses: openArray[string]): T = 
+    let client = newRpcHttpClient()
+    let router = RpcRouter.init()
+    T(rpcHttpClient: client, rpcHttpServer: newRpcHttpServer(listenAddresses, router))
+
+proc newRpcHttpProxy*(listenAddresses: openArray[string]): RpcHttpProxy =
+    RpcHttpProxy.new(listenAddresses)
+
+proc init*(proxy:RpcHttpProxy, proxyServerUrl: string) {.async.} =
```
kdeme, line 20: We use `init` for "object construction" of non ref objects, as we use `new` for ref objects, like you did correctly above. 

So I would avoid calling this `init` but instead prefer `start`

json_rpc/router.nim
```diff
@@ -71,17 +81,30 @@ proc route*(router: RpcRouter, node: JsonNode): Future[StringOfJson] {.async, gc
...
   let rpcProc = router.procs.getOrDefault(methodName)
-  if rpcProc == nil:
+  let proxyCall = router.proxyCalls.getOrDefault(methodName)
+  let params = node.getOrDefault("params")
+
+  if rpcProc == nil and (proxyCall == nil):
     return wrapError(METHOD_NOT_FOUND, "'" & methodName & "' is not a registered RPC method", id)
+  elif rpcProc == nil and (proxyCall != nil):
+    try:
+      let res = await proxyCall(methodName, if params == nil: newJArray() else: params)
```
kdeme, line 91: Regarding your question of error handling in the recovery/proxy procedure, I don't think we have a stance on this? It probably depends on the situation, but considering this is a networked (async) request and you need its return value to proceed, it probably fits better here the way you did it. (It would be different in other situations e.g. where you use `asyncSpawn`).

KonradStaniec: that was my line of thinking, that from router pov this procedure could be anything (in our case it is network call), so router needs to handle all errors this procedure can raise.

tests/testproxy.nim
```diff
@@ -0,0 +1,46 @@
...
+  test "Successful RPC call no proxy":
+    let r = waitFor client.call("myProc1", %[%"abc", %[1, 2, 3, 4]])
+    check r.getStr == "Hello abc data: [1, 2, 3, 4]"
+  test "Missing params":
+    expect(CatchableError):
+      discard waitFor client.call("myProc", %[%"abc"])
+  test "Method missing on server and proxy server":
+    expect(CatchableError):
+      discard waitFor client.call("missingMethod", %[%"abc"])
+
```
kdeme, line 41: Considering it is allowed to register a same proxy and standard handler on the server, i'd say we should add a test that tests the order of priority. That is, from current implementation the standard handler should be selected in that case.

json_rpc/rpcproxy.nim
```diff
@@ -0,0 +1,38 @@
+import
```
kdeme, line 1: For new code we put `{.push raises: [Defect].}` on top of the file: see https://status-im.github.io/nim-style-guide/errors.exceptions.html

This immediately makes you see what errors your new procedures might raise and forces you to think about that.
I realize that this module is not the best example of this and that you will mostly have to annotate the calls.

json_rpc/router.nim
```diff
@@ -12,8 +12,11 @@ type
   # Procedure signature accepted as an RPC call by server
   RpcProc* = proc(input: JsonNode): Future[StringOfJson] {.gcsafe, raises: [Defect, CatchableError].}
 
+  ProxyCall* = proc(name: string, input: JsonNode): Future[JsonNode] {.gcsafe, raises: [Defect, CatchableError].}
+
   RpcRouter* = object
     procs*: Table[string, RpcProc]
+    proxyCalls*: Table[string, ProxyCall]
```
zah, line 19: Adding this table to the router gets the job done, but it's a bit awkward specialization for such a general purpose library. A more general design is possible if you provide the users with an API allowing them to handle "raw" JSON RPC messages (i.e. one where the input is the compete `JsonNode` of the request). This can either be a callback that sits before the router in the framework or you can provide some building blocks that will make it easy for the users to implement their own reader loop handling the JSON-RPC messages one by one.

If such provisions were available, implementing the proxy would be a matter of gluing together some of the lower-level components in the user application (or in a helper module such as `rpcproxy`).

KonradStaniec: I kinda agree with you about this specialisation issue (or maybe it is just the question of naming, maybe it should be names `MissingMethodHandler` ).

My point is that I see router as this kinda glue layer, which enables user of the libary to connect server with other stuff, and `rpcproxy` as it currently stands just glues http server with client + some methods to make everything more easy to use, and I am not sure that adding another level of indirection is necessary here.

My another approach was to make router polymorphic i.e have some base class with some of the basic methods and one method to handle `JsonNode` of the request which would need to be overriden in subclasses. Then we could have basic router as it is now, we could have `ProxyRouter` which could have this addtional table, and user of library could implement its own router with whatever abilities he wants. But I ultimately deemed this approach to too complex for current needs. 

Either way let my know what you think about polymorphic approach and I will try to experiment with approach you suggested.

zah: Indeed, calling this a `MissingMethodHandler` probably hits the sweet spot of being general purpose and practical enough for inclusion in the library. But then, it would make more sense to route all missing methods to the handler. You wouldn't have these lookups by method name (for the proxy calls).

KonradStaniec: Having it in the table solves some other problems:
- user of router needs deliberately decide which methods will use missing method handler and which will use normal rpc call, which disables possibility of proxying some unwanted methods by accident. (as all other methods will return missing method error)
- user can in theory have different handlers for different calls (so lets if some endpoints require addtional handling, formating , logging etc)

zah: For me, a `MissingMethodHandler` will be responsible for handling all requests that doesn't have other handlers installed. Some people will be familiar with the concept form Ruby's [method_missing](https://apidock.com/ruby/BasicObject/method_missing) feature which is consistent with such a behavior. 

Otherwise, what's the difference between a RPC handler and a proxy handler? Instead of having a separate proxy table, the regular RPC handlers can accept a `JsonNode` argument which would skip the automated parameter marshaling logic. This, combined with the ability to make "raw" requests by passing in a `JsonNode` in `client.call(name, params)`, would be another way to implement a proxy.

KonradStaniec: you are totally right ! My confusion was from the fact that I missed when the serialization/deserialization was really taking place, but as you mention current rpc proc (not macro) deals with raw request, so it is entirely possible to implement proxy by using current router capabilities by just having partially applied proc which returns correct `RpcProc` type. Thanks for discussion :+1:

so in summary:
- I have removed the additional table as it is not necessary
- I have left proxy implementation as it nice place to glue all the code
- I have removed missing handler as it is not neccessary to proxy implementation, if you think it would be valuable addition I can add it in spearate pr.

PR #134, opened 2022-04-06
Facilitate http code response on authentication failure

mjfh:


json_rpc/servers/websocketserver.nim
```diff
@@ -10,18 +10,51 @@ logScope:
...
+    authHook: Option[RpcWebSocketServerAuth] ## Authorization call back handler
     server: StreamServer
     wsserver: WSServer
 
+  HookEx = ref object of Hook
+    handler: RpcWebSocketServerAuth ## from `RpcWebSocketServer`
+    request: HttpRequest            ## current request needed for error response
+
+proc authWithHtCodeResponse(ctx: Hook, headers: HttpTable):
+            Future[Result[void, string]] {.async, gcsafe, raises: [Defect].} =
```
arnetheduck, line 30: async functions should not have direct raises annotations - gcsafe should also not be needed, why is it here?

PR #136, opened 2022-04-10
Fix connection leaks in the HTTP client

zah:


json_rpc/clients/httpclient.nim
```diff
@@ -65,17 +65,38 @@ method call*(client: RpcHttpClient, name: string,
...
-        await req.send()
-      except CancelledError as e:
-        raise e
-      except CatchableError as e:
-        raise (ref RpcPostError)(msg: "Failed to send POST Request with JSON-RPC.", parent: e)
+
+  var req: HttpClientRequestRef
+  var res: HttpClientResponseRef
+
+  defer:
```
arnetheduck, line 72: @cheatfate after `req.send`, is there anything else but `close` that you can do with a request?

json_rpc/clients/httpclient.nim
```diff
@@ -65,17 +65,38 @@ method call*(client: RpcHttpClient, name: string,
...
+    # Using multiple defer statements in this function or multiple
+    # try/except blocks within a single defer statement doesn't
+    # produce the desired run-time code, so we use slightly bizzare
+    # code to ensure the exceptions safety of this function:
+    try:
+      var closeFutures = newSeq[Future[void]]()
+      if req != nil: closeFutures.add req.closeWait()
+      if res != nil: closeFutures.add res.closeWait()
+      if closeFutures.len > 0: await allFutures(closeFutures)
+    except CatchableError as err:
```
arnetheduck, line 83: exceptions in finally handlers are problematic in general - does `closeWait` really raise? if so, I suspect that needs fixing so that it never raises, or it becomes quite difficult to reason about shutdown in general

json_rpc/clients/httpclient.nim
```diff
@@ -65,17 +65,38 @@ method call*(client: RpcHttpClient, name: string,
...
+    # BEWARE!
+    # Using multiple defer statements in this function or multiple
+    # try/except blocks within a single defer statement doesn't
+    # produce the desired run-time code, so we use slightly bizzare
+    # code to ensure the exceptions safety of this function:
+    try:
+      var closeFutures = newSeq[Future[void]]()
+      if req != nil: closeFutures.add req.closeWait()
+      if res != nil: closeFutures.add res.closeWait()
+      if closeFutures.len > 0: await allFutures(closeFutures)
```
kdeme, line 82: From testing on Fluffy, the waiting here causes quite a visible slow down when doing many sequential json-rpc requests.

PR #146, opened 2022-09-01
align `nim-websock` support with `news`

etan-status:
To allow switching between `news` and `nim-websock` in `nimbus-eth2`,
add support for the missing CORS and JWT authorization headers to the
`nim-websock` backend. Also transition from `StringTableRef` to
`HttpTable` to properly support multiple instances of same header name.

### DIFF
--- a/json_rpc/clients/websocketclient.nim
+++ b/json_rpc/clients/websocketclient.nim
@@ -1,6 +1,5 @@
 import
-  std/[strtabs, tables],
-  pkg/[chronos, chronicles],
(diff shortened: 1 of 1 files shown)


etan-status:
macOS nim devel is due to `could not load: libcrypto.1.1.dylib` which is a temporary issue where `brew install openssl@1.1` is required. It's unrelated to this PR.

etan-status:
Tested both backends as part of `nimbus-eth2` using Goerli sync

PR #149, opened 2022-09-15
Safer async usage in httpclient.call

Menduist:


Menduist:
#150 was merged instead

PR #151, opened 2022-09-15
Use JsonRpcError instead of ValueError

emizzle:
Allows implementors of this library to distinguish between errors originating from this library.

Likely a breaking change for current users.

### DIFF
--- a/json_rpc/client.nim
+++ b/json_rpc/client.nim
@@ -1,6 +1,7 @@
 import
   std/[tables, macros],
   chronos,
+  ./errors,
   ./jsonmarshal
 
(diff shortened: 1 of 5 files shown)


arnetheduck:
There are a _a lot_ more random errors coming out from practically any call in here (chronos, json parsing etc) - will need to do be a lot more thorough if you want to narrow down the exceptions in here - ie practically every line in the library needs to live in a try/catch with an error conversion

arnetheduck:
@emizzle this is still a valuable change - will you be getting back to it?

jangko:
close in favor of #178

PR #154, opened 2022-10-21
expose http server options

arnetheduck:
These are needed to create a server that can handle larger POST requests


arnetheduck:
Nim devel test fail is outside the scope of this PR

PR #160, opened 2023-02-26
Fix Nim 1.6 regressions

fox0430:
Ref: https://github.com/status-im/nim-unittest2/issues/23


fox0430:
Any response? nim-json-rpc cannot compile with Nim 1.6. Does the official compiler not support it?

zah:
Thank you! And sorry for the slow response

fox0430:
@zah No problem! Thank you.

PR #161, opened 2023-04-27
Remove nimble.lock file to be able to run nimble test

kdeme:
With the nimble.lock file it is not possible to run any nimble commands from within a nimbus-build-system project env.

Not sure what the deal is with this and if there is another way to fix it currently.


ringabout:
Hey @arnetheduck could it be merged (at least restart CI; I checked before on my fork, which passed the devel testss), so `nim-json-rpc` can be added to important packages?

ringabout:
Thanks!

PR #178, opened 2024-01-03
Upgrade rpc router internals

### LINKED ISSUE #20: Exception handling
Use more appropriate exception objects rather than just ValueError.
Review unnecessary exceptions and where appropriate convert to error codes.

### LINKED ISSUE #90: Feature request: named parameters for Request Objects
Currently the router requires parameters for rpc calls to be specified as a `JArray`. This implies that params can only be provided "by-position". The [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification#parameter_structures) also specifies providing parameters "by-name" in an `Object`.
This will be a useful feature in nim-json-rpc and improve readability of requests.

### LINKED ISSUE #128: Undocumented compilation error using `params` as param name in rpc method definition
The following rpc method definition using `params` as the param name will trigger compilation error. If renamed to something else it will be ok. looks like the internal templates and macros are messing with it because they use `params` as their working variable. If we cannot fix it, then doc it at least.

```Nim
server.rpc("method1") do(params: string) -> bool:
   return true
```

Error message:
```text
...\engine_api.nim(83, 9) template/generic instantiation of `rpc` from here
...\json_rpc\server.nim(19, 16) template/generic instantiation of `rpc` from here
...\json_rpc\router.nim(155, 73) template/generic instantiation of `async` from here
...\json_rpc\jsonmarshal.nim(244, 22) Error: undeclared field: 'elems' for type system.string [declared in F:\projects\status-nim\lib\system.nim(34, 3)]
```

jangko:
fix #128
fix #90
fix #151
fix #20 

And also completing migration to json-serialization


json_rpc/clients/httpclient.nim
```diff
@@ -84,6 +87,7 @@ method call*(client: RpcHttpClient, name: string,
         await res.closeWait()
       except CatchableError as exc: # shouldn't happen
         debug "Error closing JSON-RPC HTTP resuest/response", err = exc.msg
+    {.hint[XDeclaredButNotUsed]:on.}
```
arnetheduck, line 90: this turns it on even if it was disabled before.. why not use `{.used.}`?

json_rpc/client.nim
```diff
@@ -8,195 +8,123 @@
...
-      result.add(procDef)
-
-proc setMethodHandler*(cl: RpcClient, name: string, callback: MethodHandler) =
-  cl.methodHandlers[name] = callback
-
-proc delMethodHandler*(cl: RpcClient, name: string) =
-  cl.methodHandlers.del(name)
+  except CatchableError as exc:
+    return err(exc.msg)
+  except Exception as exc:
```
arnetheduck, line 99: where is this Exception coming from?

json_rpc/private/jrpc_conv.nim
```diff
@@ -0,0 +1,94 @@
...
+  value = json_serialization.parseInt(r, uint8).byte
+
+proc writeValue*(w: var JsonWriter[JrpcConv], value: byte) =
+  json_serialization.writeValue(w, uint8(value))
+
+# Enums.
+
+proc readValue*(r: var JsonReader[JrpcConv], value: var (enum)) =
+  wrapErrors r, value:
+    value = type(value) json_serialization.readValue(r, uint64)
```
arnetheduck, line 55: this looks like it will defect on non-members of the enum (or worse)

json_rpc/private/jrpc_conv.nim
```diff
@@ -0,0 +1,94 @@
...
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import
+  std/[tables, macros],
+  results,
+  json_serialization,
+  json_serialization/stew/results as jser_results,
```
arnetheduck, line 14: results is no longer part of stew

json_rpc/private/jrpc_conv.nim
```diff
@@ -0,0 +1,94 @@
...
+  sets,
+  tables,
+  json_serialization,
+  jser_results
+
+type
+  StringOfJson* = JsonString
+
+createJsonFlavor JrpcConv,
+  automaticObjectSerialization = true,
```
arnetheduck, line 28: in nimbus we've had several cases of objects not intended for serialization end up getting caught in these generic open-for-all serializers - this is an edge case because _technically_ this serializer looks like it's only used during parameter conversion which in a way is an explicit choice - nonetheless, having it accept any object seems risky.

json_rpc/private/jrpc_conv.nim
```diff
@@ -0,0 +1,94 @@
...
+# at your option.
+# This file may not be copied, modified, or distributed except according to
+# those terms.
+
+import
+  std/[tables, macros],
+  results,
+  json_serialization,
+  json_serialization/stew/results as jser_results,
+  json_serialization/std/[options, sets, tables]
```
arnetheduck, line 15: why are we import/exporting these? it seems like an application choice?

json_rpc/private/jrpc_sys.nim
```diff
@@ -0,0 +1,350 @@
...
+  of JsonValueKind.Array:
+    val = RequestBatchRx(kind: rbkMany)
+    r.readValue(val.many)
+  of JsonValueKind.Object:
+    val = RequestBatchRx(kind: rbkSingle)
+    r.readValue(val.single)
+  else:
+    r.raiseUnexpectedValue("RequestBatch must be either array or object, got=" & $tok)
+
+proc writeValue*[F](w: var JsonWriter[F], val: RequestBatchRx) =
```
arnetheduck, line 305: ditto on object safety - it seems better to use an opt-in serializer here, also for future refactoring when other types that shouldn't be serialized accidentally might end up getting matched.

json_rpc/clients/httpclient.nim
jangko: fixed/removed

json_rpc/client.nim
jangko: removed

json_rpc/private/jrpc_conv.nim
jangko: deleted, they are remnants of previous attempts

json_rpc/private/jrpc_conv.nim
jangko: fixed

json_rpc/private/jrpc_conv.nim
jangko: removed

json_rpc/private/jrpc_sys.nim
jangko: fixed, they are now using custom serializer.

json_rpc/private/jrpc_conv.nim
jangko: removed

jangko:
Thank you for your review @arnetheduck, I've remove all unsafe serializer.  Now the user have to explicitly register their types or custom serializer for json rpc to work.

json_rpc/private/jrpc_sys.nim
```diff
@@ -0,0 +1,353 @@
...
+  let version = r.parseAsString()
+  if version != JsonRPC2Literal:
+    r.raiseUnexpectedValue("Invalid JSON-RPC version, want=" &
+      JsonRPC2Literal.string & " got=" & version.string)
+
+proc writeValue*(w: var JsonWriter[JrpcSys], val: JsonRPC2)
+      {.gcsafe, raises: [IOError].} =
+  w.writeValue JsonRPC2Literal
+
+proc readValue*(r: var JsonReader[JrpcSys], val: var ResponseError)
```
arnetheduck, line 186: if we disable automatic object serialization in JrcpSys and then opt in for these few types, I think we can avoid these manual case implementations?

jangko: Right, it's too easy to write custom serializer using the new json-serialization, make me forget to use new features in nim-serialization. Now it's fixed.

json_rpc/private/jrpc_sys.nim
```diff
@@ -142,6 +143,11 @@ type
 createJsonFlavor JrpcSys,
   requireAllFields = false
 
+ResponseError.useDefaultSerializationIn JrpcSys
```
arnetheduck, line 146: or `JrpcSys.useDefaultSerializationFor(ResponseError, RequestTx, ...)`, right?

My preferred option would actually be `type RequestTx {.json: [JrpcSys].}` or something along those lines that but I suspect we might run into macro constraints implementing that..

json_rpc/private/jrpc_sys.nim
arnetheduck: :rofl:

json_rpc/private/jrpc_sys.nim
jangko: > JrpcSys.useDefaultSerializationFor(ResponseError, RequestTx, ...), right?

No, because only `ResponseError` using both reader dan writer, while the rest use only reader or writer, not both.

jangko: > My preferred option would actually be type RequestTx {.json: [JrpcSys].} or something along those lines that but I suspect we might run into macro constraints implementing that..

I think it's doable. We can use simple template custom pragma instead of macro. The macro part already in nim-serialization. So it's still fall within nim-serialization capabilities.

PR #184, opened 2024-01-04
Add test for createRpcSigsFromNim and createSingleRpcSig

jangko:


tests/test_callsigs.nim
```diff
@@ -8,19 +8,127 @@
...
+      return true
+    return false
+
+  s.rpc("getFilter") do(id: string) -> string:
+    if id == "cow":
+      return "moo"
+    return "meow"
+
+suite "test callsigs":
+  var server = newRpcSocketServer(["127.0.0.1:8545"])
```
arnetheduck, line 82: better to use port 0 in tests to let the OS choose

PR #196, opened 2024-01-21
Client also handle error message if id is null

jangko:


json_rpc/clients/httpclient.nim
```diff
@@ -152,6 +152,7 @@ method call*(client: RpcHttpClient, name: string,
   if msgRes.isErr:
     # Need to clean up in case the answer was invalid
     debug "Failed to process POST Response for JSON-RPC", msg = msgRes.error
+    newFut.cancelSoon()
```
arnetheduck, line 155: shouldn't this pass a more useful error to newFut?

jangko: Last time I checked, the raise statement below still produce adequate error. But I think I'll learn how to use chronos features.

arnetheduck: well, there are two futures involved in this snippet - one is the  `call` future returned from the method - the other is the `newFut` that presumably something else is waiting for? The raise below is for the former while `newFut` needs its own treatment.

jangko: Ah, okay. Thanks

PR #216, opened 2024-03-01
Introduce custom application error with optional data object

kdeme:
This addresses two items:
- Use ApplicationError instead of InvalidRequest for custom application errors. Avoiding confusing regarding the usage of InvalidRequest (although the actual error code used could/would be different).
- Allow for defining an optional data object that gets returned in the error response, as per json-rpc specification.

### DIFF
--- a/json_rpc/errors.nim
+++ b/json_rpc/errors.nim
@@ -1,11 +1,15 @@
 # json-rpc
-# Copyright (c) 2019-2023 Status Research & Development GmbH
+# Copyright (c) 2019-2024 Status Research & Development GmbH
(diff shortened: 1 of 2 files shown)


jangko (approved):

jangko:
I think this is a good idea.

kdeme:
I'm not fully sure yet about keeping all other exceptions (the `CatchableError` except clause) as system errors, but I'll leave it as that for now.

It might be nice to evolve in the future to a `ResponseError` that would be returned by the actual handlers (instead of exceptions).

PR #217, opened 2024-03-04
chore: changing log level to trace for sending message print

gabrielmer:
In applications with recurring RPC requests, this print can pollute the logs way too much. 

As a result, propose to have this log set to trace :)

### DIFF
--- a/json_rpc/clients/httpclient.nim
+++ b/json_rpc/clients/httpclient.nim
@@ -142,7 +142,7 @@ method call*(client: RpcHttpClient, name: string,
     id = client.getNextId()
     reqBody = requestTxEncode(name, params, id)
 
-  debug "Sending message to RPC server",
+  trace "Sending message to RPC server",
          address = client.httpAddress, msg_len = len(reqBody), name
   trace "Message", msg = reqBody
 


Ivansete-status (approved):
LGTM! Thanks!

gabrielmer:
@etan-status @tersec WDYT?

json_rpc/clients/httpclient.nim
```diff
@@ -142,7 +142,7 @@ method call*(client: RpcHttpClient, name: string,
     id = client.getNextId()
     reqBody = requestTxEncode(name, params, id)
 
-  debug "Sending message to RPC server",
+  trace "Sending message to RPC server",
```
arnetheduck, line 145: broadly, this logging is currently needed for nimbus - what could be done is to implement a hook that would be called before/after each request - in such a world, we could remove logging entirely from json-rpc (notably, logging is used when the API does not expose enough information for the consumer of the api to log it by themselves - it's a shortcut to avoid complicating the API)

tersec:
It's already `debug`, so someone opting into that log level is already opting into a fairly sizable volume of logs.

gabrielmer:
Thanks for your input! The hooks approach is actually great, but it's true that if it's in `debug` we're already opting for extensive logging. So will close the PR so we can keep it simple :)

Apologies and thank you!

PR #220, opened 2024-06-18
Add redefine pragma to inner template

jangko:
A nim 2.0 feature


arnetheduck:
need to keep 1.6 compat for a good while, in these core libraries

jangko:
This changes still support nim 1.6

PR #222, opened 2024-07-17
Some minimal changes to be able to base the `nimlangserver` in `json_rpc`

jmgomez:
- `processClientHook`: This is useful to shape the request and response to defined protocols (i.e. lsp)
- overloads `tryRoute`: So it doesnt parse the `json` inside.

### DIFF
--- a/json_rpc/router.nim
+++ b/json_rpc/router.nim
@@ -196,17 +196,15 @@ proc route*(router: RpcRouter, data: string):
 
   return reply
 
-proc tryRoute*(router: RpcRouter, data: JsonString,
+proc tryRoute*(router: RpcRouter, req: RequestRx,
                fut: var Future[JsonString]): Result[void, string] =
   ## Route to RPC, returns false if the method or params cannot be found.
-  ## Expects json input and returns json output.
+  ## Expects RequestRx input and returns json output.
   when defined(nimHasWarnBareExcept):
     {.push warning[BareExcept]:off.}
     {.push warning[UnreachableCode]:off.}
 
   try:
-    let req = JrpcSys.decode(data.string, RequestRx)
-
     if req.jsonrpc.isNone:
       return err("`jsonrpc` missing or invalid")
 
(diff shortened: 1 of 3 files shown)


jmgomez:
Failures seem unrelated cc @arnetheduck

tests/testserverclient.nim
```diff
@@ -114,3 +114,22 @@ suite "Websocket Server/Client RPC with Compression":
...
 
+suite "Custom processClient":
+  test "Should be able to use custom processClient":
+    var srv = newRpcSocketServer()
+    var wasCalled = false
+    
+    proc processClientHook(server: StreamServer, transport: StreamTransport) {.async: (raises: []), gcsafe.} =
+      wasCalled = true
+    
+    srv.processClientHook = processClientHook
```
jangko, line 125: Better to add this assignment in constructor

json_rpc/router.nim
```diff
@@ -229,6 +227,15 @@ proc tryRoute*(router: RpcRouter, data: JsonString,
     {.pop warning[BareExcept]:on.}
     {.pop warning[UnreachableCode]:on.}
 
+proc tryRoute*(router: RpcRouter, data: JsonString,
+               fut: var Future[JsonString]): Result[void, string] =
+  ## Route to RPC, returns false if the method or params cannot be found.
+  ## Expects json input and returns json output.
+  try:
+    let req = JrpcSys.decode(data.string, RequestRx)
```
jangko, line 235: Why the original `tryRoute` not called here? That is exactly what caused the failures.

jangko (changes requested):

json_rpc/servers/socketserver.nim
```diff
@@ -18,6 +18,7 @@ export errors, server
 type
   RpcSocketServer* = ref object of RpcServer
     servers: seq[StreamServer]
+    processClientHook*: StreamCallback2
```
jangko, line 21: There is no reason to expose this field to the public.

jmgomez: Will add it to the constructor and remove the `*`. The reasoning was to follow what `StreamServer` does.

json_rpc/router.nim
jmgomez: Ah right, a diff issue. Sorry about that

jangko (approved):

PR #224, opened 2024-10-03
Implement isConnected method

bhartnett:
Having an isConnected method on the RpcClient will be useful for handling reconnection scenarios especially for websockets. At the moment if the websocket server goes offline then restarts the client is disconnected and not reconnected. Having an isConnected proc will allow checking and reconnecting in applications.

### DIFF
--- a/json_rpc/clients/httpclient.nim
+++ b/json_rpc/clients/httpclient.nim
@@ -220,8 +220,12 @@ proc connect*(client: RpcHttpClient, address: string, port: Port, secure: bool) 
   else:
     raise newException(RpcAddressUnresolvableError, res.error)
 
+method isConnected*(client: RpcHttpClient): bool =
+  (not client.httpSession.isNil()) and client.httpAddress.isOk()
+
 method close*(client: RpcHttpClient) {.async.} =
   if not client.httpSession.isNil:
     await client.httpSession.closeWait()
+    client.httpSession = nil
 
 {.pop.}
--- a/tests/testserverclient.nim
+++ b/tests/testserverclient.nim
@@ -49,7 +49,49 @@ suite "Socket Server/Client RPC":
     except CatchableError as e:
       check e.msg == """{"code":-32001,"message":"Unknown payload"}"""
 
+  test "Client close and isConnected":
+    check client.isConnected() == true
+    # Is socket server close broken?
+    # waitFor client.close()
+    # check client.isConnected() == false
+
   srv.stop()
+  waitFor srv.closeWait()
+
+suite "HTTP Server/Client RPC":
+  var srv = newRpcHttpServer([initTAddress("127.0.0.1", Port(0))])
+  var client = newRpcHttpClient()
+
+  echo "address: ", $srv.localAddress()
+  srv.setupServer()
+  srv.start()
+  waitFor client.connect("http://" & $(srv.localAddress()[0]))
+
+  test "Successful RPC call":
+    let r = waitFor client.call("myProc", %[%"abc", %[1, 2, 3, 4]])
+    check r.string == "\"Hello abc data: [1, 2, 3, 4]\""
+
+  test "Missing params":
+    expect(CatchableError):
+      discard waitFor client.call("myProc", %[%"abc"])
+
+  test "Error RPC call":
+    expect(CatchableError): # The error type wont be translated
+      discard waitFor client.call("myError", %[%"abc", %[1, 2, 3, 4]])
+
+  test "Invalid request exception":
+    try:
+      discard waitFor client.call("invalidRequest", %[])
+      check false
+    except CatchableError as e:
+      check e.msg == """{"code":-32001,"message":"Unknown payload"}"""
+
+  test "Client close and isConnected":
+    check client.isConnected() == true
+    waitFor client.close()
+    check client.isConnected() == false
+
+  waitFor srv.stop()
   waitFor srv.closeWait()
 
 suite "Websocket Server/Client RPC":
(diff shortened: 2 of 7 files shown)


json_rpc/clients/httpclient.nim
```diff
@@ -220,8 +220,12 @@ proc connect*(client: RpcHttpClient, address: string, port: Port, secure: bool)
   else:
     raise newException(RpcAddressUnresolvableError, res.error)
 
+method isConnected*(client: RpcHttpClient): bool =
+  (not client.httpSession.isNil()) and client.httpAddress.isOk()
+
 method close*(client: RpcHttpClient) {.async.} =
   if not client.httpSession.isNil:
     await client.httpSession.closeWait()
+    client.httpSession = nil
```
bhartnett, line 229: It appears that this change has broken the CI for some reason. I'm looking into the cause.

jangko: testhttps.nim too 
https://github.com/status-im/nim-json-rpc/blob/9da5b609bf27805ffaa598efd2db02c1eb5019f4/tests/testhttps.nim#L91

jangko:
RpcHttpClient actually not using persistent connection like RpcWebsocketClient or RpcSocketClient. RpcHttpClient does not using keep-alive connection. Every request actually a new connection. I don't think we need to add isConnected to RpcHttpClient.

bhartnett:
> RpcHttpClient actually not using persistent connection like RpcWebsocketClient or RpcSocketClient. RpcHttpClient does not using keep-alive connection. Every request actually a new connection. I don't think we need to add isConnected to RpcHttpClient.

I see, in that case it will be unimplemented and throw an assertion Defect if called on the RpcClient type. That might create a problem. For example, I've created a type that uses RpcClient and it might be a RpcHttpClient or a RpcWebsocketClient and I'll like to check the connection if an error is thrown in these cases with having to right code that checks what type of instance it is. 

Perhaps it would be better to simply return true for the RpcHttpClient? The RpcHttpClient does have a connect method already so perhaps isConnected in this context just means that it is ready to send requests.

tests/testserverclient.nim
```diff
@@ -49,9 +49,51 @@ suite "Socket Server/Client RPC":
     except CatchableError as e:
       check e.msg == """{"code":-32001,"message":"Unknown payload"}"""
 
+  test "Client close and isConnected":
+    check client.isConnected() == true
+    # Is socket server close broken?
+    # waitFor client.close()
```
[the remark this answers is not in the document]
bhartnett: I didn't spend long looking at it. Do you have a suggestion/guidance on how I can fix it? If it isn't trivial, then I'll create an issue for it.

json_rpc/clients/httpclient.nim
bhartnett: I see, thanks. I'll fix.

jangko:
> Perhaps it would be better to simply return true for the RpcHttpClient?

I have no objection, although it's a bit misleading technically. But I think it's ok. `isConnectedOrReadyToSend` is too wordy.

bhartnett:
> > Perhaps it would be better to simply return true for the RpcHttpClient?
> 
> I have no objection, although it's a bit misleading technically. But I think it's ok. `isConnectedOrReadyToSend` is too wordy.

Yes thats true but perhaps if we implement keep alive then it won't matter. In that case, I'll leave it for now.

tests/testserverclient.nim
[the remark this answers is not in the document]
bhartnett: I've just pushed a fix which solves the problem in the tests. Can you review the change when you get a chance? I'm not sure if it might effect something else.

jangko (approved):

bhartnett:
> > Having an isConnected proc will allow checking and reconnecting in applications.
> 
> this is based on the assumption that connectivity accurately can be established, which is usually not true - ie the normal case is that you discover if you're connected or not on the first read or write from the socket, so the right place to handle reconnections is _usually_ in one of those two places: where you wait for new data to arrive and where you attempt to send stuff.

What would you suggest for this PR in that case? The overall goal for this PR was to have a way to decide whether to try reconnecting when using websockets and catching an error after issuing a json-rpc request. If the web socket connection is lost then a reconnect is required in order to continue sending data. 

In the RpcWebSocketClient type the WSSession transport field gets set to nil when the connection is lost and as you said this is only discovered when sending or receiving. Currently the only way for me to handle this is to check if the wsClient.transport field is nil which doesn't seem like a very nice solution.

arnetheduck:
> What would you suggest for this PR in that case?

The "broad" way of handling this is to have the thing doing the requesting decide if it should retry based on the outcome of the request itself - the way sockets work is that many connection failures are detected when you try to use the socket, so this is usually the best place to "schedule" retries.

Wanting to check for `isConnected` _usually_ indicates that there is some underlying architectural issue, for example that it is not clear who "owns" the responsibility to open and close the connection when a request is being made.

bhartnett:
> > What would you suggest for this PR in that case?
> 
> The "broad" way of handling this is to have the thing doing the requesting decide if it should retry based on the outcome of the request itself - the way sockets work is that many connection failures are detected when you try to use the socket, so this is usually the best place to "schedule" retries.
> 
> Wanting to check for `isConnected` _usually_ indicates that there is some underlying architectural issue, for example that it is not clear who "owns" the responsibility to open and close the connection when a request is being made.

It sounds like a better design for this might be to simply throw an exception at the time of doing a request which clearly indicates that a reconnect is required. The calling code can then catch and then handle appropriately. I'll close this PR for now and come up with a better solution.

PR #233, opened 2024-11-28
use the common CI workflow

narimiran:


jangko:
It's not that this library not compatible with Nim 1.6 anymore but a problem in nim-websock. The problem have been fixed in nim-websock but tagged properly. I have fixed that too. Can you try once again with Nim 1.6 enabled?

narimiran:
> Can you try once again with Nim 1.6 enabled?

It seems the problem is still here.

Do I need to rebase this onto the latest commit in the master branch, or something else needs to be done?

PR #235, opened 2025-01-22
Update httpclient.nim move log to trace

Ivansete-status:
In nwaku, when the node's user sends a request to their infura RPC, the token is being printed by default.

This PR aims to avoid printing it by default and just if someone is interested in debugging any possible issue

### DIFF
--- a/json_rpc/clients/httpclient.nim
+++ b/json_rpc/clients/httpclient.nim
@@ -145,7 +145,7 @@ method call*(client: RpcHttpClient, name: string,
     id = client.getNextId()
     reqBody = requestTxEncode(name, params, id)
 
-  debug "Sending JSON-RPC request",
+  trace "Sending JSON-RPC request",
          address = client.httpAddress, len = len(reqBody), name, id
   trace "Message", msg = reqBody
 
--- a/json_rpc/clients/httpclient.nim
+++ b/json_rpc/clients/httpclient.nim
@@ -181,7 +181,7 @@ method callBatch*(client: RpcHttpClient,
                   calls: RequestBatchTx): Future[ResponseBatchRx]
                     {.async.} =
   let reqBody = requestBatchEncode(calls)
-  debug "Sending JSON-RPC batch",
+  trace "Sending JSON-RPC batch",
         address = $client.httpAddress, len = len(reqBody)
   let resText = await client.callImpl(reqBody)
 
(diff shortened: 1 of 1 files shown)


vpavlin (dismissed):
~LGTM, this should also help with spammy output of this lib in nwaku and codex in debug mode~

Actually, @Ivansete-status would you mind going through the other files as well? The address is being logged again a few lines below for batch request and in websocket clients etc.

It would be probably good to not solve this ad-hoc of this one particular line, but make it in the whole lib.

WDYT?

chaitanyaprem (commented):
wouldn't the token get logged if trace level is enabled which i think is possible?

vpavlin:
It would, but if the goal of the library is to log the address, we cannot prevent logging the token which is part of the address (unless some obfuscation can be done). Not sure what a more reasonable approach than moving it to `trace` would be

arnetheduck:
A similar question was asked here: https://github.com/status-im/nim-chronicles/issues/157 - see the suggested solutions - in particular, the type-based solution might be more reasonable here, ie one could create a wrapper type that obfuscates url parameters or something like that, so that it gets applied consistently across all logging levels.

Changing level doesn't really solve the problem.

arnetheduck (changes requested):
..

Ivansete-status:
@arnetheduck - Should we then implement the suggestion in `chronicles` instead?

arnetheduck:
> Should we then implement the suggestion in chronicles instead?

what exactly would you implement in chronicles?

ie the way I see it, a `PrivateUrl` type would be appropriate, and its printing helpers would simply mask out sensitive fields.

arnetheduck:
.. or whatever type is currently used when logging, just add a `formatIt` to that type...

Ivansete-status:
Thanks so much for the comments @arnetheduck !
Let me know if it looks better/simpler: https://github.com/status-im/nim-json-rpc/pull/235/commits/927604568ea6457eaa2c4f3d2b7ea9e9dd7e386d

Ivansete-status:
Apologies for the noise guys!

This is already well handled and in fact, in my local run it always prints the host+port.
The user might had been running an old nwaku version.

This is the commit that enhanced the log by only showing `id`:
https://github.com/status-im/nim-json-rpc/commit/98a5efba4de26ac852d0715656f6b0c52a203a75

image

... being `id` created in the following line:

https://github.com/status-im/nim-chronos/blob/c04576d829b8a0a1b12baaa8bc92037501b3a4a0/chronos/apps/http/httpclient.nim#L424

Closing this PR as it is not needed at all

PR #240, opened 2025-06-28
Add logScope topics to router module

jangko:


json_rpc.nimble
```diff
@@ -17,7 +17,7 @@ license       = "Apache License 2.0"
 skipDirs      = @["tests"]
 
 ### Dependencies
-requires "nim >= 1.6.0",
+requires "nim >= 2.0.16",
```
arnetheduck, line 20: the taskrequires thing is a bug in chronicles, ie we should remove it from there

arnetheduck: https://github.com/status-im/nim-chronicles/commit/a537556963bb42e048c7f7229456995355d360e5#r160993642

basically, when compiling with nim 1.6, nimble should choose a version of chronicles that is compatible with 1.6 but it doesn't because it first checks 0.11 which it cannot parse and crashes .. I *think* adding back the `when` in chronicles should fix this (we might have to move the 0.11 tag).. let's not make the version bump in nim-json-rpc, it's an interesting problem to fix in nimble

arnetheduck: hm, bummer - the nimble that ships with nim 1.6 can't deal with this - the newer nimble can however ..

arnetheduck: https://github.com/status-im/nim-web3/pull/211 fixes enough nimble problems that a new nimble can work with this setup

PR #263, opened 2026-02-22
Documentation book

nitely:
New docs preview: https://nitely.github.io/nim-json-rpc/

Structure based on [vs-streamjsonrpc](https://microsoft.github.io/vs-streamjsonrpc/docs/getting-started.html) docs.

Changes:

- Adds mdbook
- Adds some missing exports in server/socketsserver.nim
- Update readme to link to the docs and remove the rest of usage docs.


docs/src/connecting.md
```diff
@@ -0,0 +1,79 @@
...
+- Sockets and pipes, via [chronos](https://github.com/status-im/nim-chronos)' `StreamTransport`: bidirectional, persistent connection, custom message framing.
+  - `Framing.httpHeader`: `Content-Length` prefix specifying the length of the payload, compatible with [vscode-jsonrpc](https://www.npmjs.com/package/vscode-jsonrpc).
+  - `Framing.lengthHeaderBE32`: Big-endian, 32-bit binary prefix - most efficient option.
+- Websockets: bidirectional, persistent connection.
+
+## Server (and possibly client also)
+
+Create the server instance using one of the available transports:
+
+HTTP:
```
arnetheduck, line 17: hm, for nimbus.guide we have these nice tabs: 

Image

maybe something like this is avail for mdbook as well?

docs/src/receiving_requests.md
```diff
@@ -0,0 +1,58 @@
...
+
+When a request is received, the router matches it to a server method that was previously registered with a matching name. If no matching server method can be found the request is dropped, and an error is returned to the client if the client requested a response.
+
+When an RPC-invoked server method throws an exception, the server will handle the exception and (when applicable) send an error response to the client with a description of the failure.
+
+JSON-RPC is an inherently asynchronous protocol. Multiple concurrent requests are allowed. Methods are invoked as the requests are processed, even while prior requests are still running.
+
+## Registering methods
+
+The `rpc` macro accepts a list of proc definitions which are turned into async procedures and registered as RPC methods. Procedure overload is not supported. A format flavor supporting the parameters and return type must be set. The `RpcConv` defined in the [flavors section](./format_conversion.md) is used in the following example:
```
arnetheduck: the concept of "format flavors" needs to be introduced before it's used here

docs/src/connecting.md
```diff
@@ -0,0 +1,79 @@
...
+{{#shiftinclude auto:../examples/socket_client.nim:ClientConnect}}
+```
+
+Websockets:
+
+```nim
+{{#shiftinclude auto:../examples/websocket_client.nim:ClientConnect}}
+```
+
+You can then [proceed to send requests](./sending_requests.md).
```
arnetheduck, line 65: the flow is somewhat confusing here - for the server, there is no such link and the concept of peer-to-peer vs server/client is lost here, ie this page refers to the establishment of communication but since channels can be bidirectional, it ends there - it might be worth calling this point out even though it's been introduced in the overview

docs/src/receiving_requests.md
```diff
@@ -0,0 +1,58 @@
+# Receiving a JSON-RPC request
+
+Before receiving any request, you should have already [established a connection](./connecting.md).
+
+When a request is received, the router matches it to a server method that was previously registered with a matching name. If no matching server method can be found the request is dropped, and an error is returned to the client if the client requested a response.
```
arnetheduck:
```suggestion
When a request is received, the router matches it to a server method by name. If no matching server method can be found the request is dropped, and an error is returned to the client if the client requested a response.
```

docs/src/overview.md
```diff
@@ -0,0 +1,19 @@
...
+
+JSON-RPC is a two-party, peer-to-peer based protocol by which one party can request a method's invocation of the other party, and optionally receive a response from the server with the result of that invocation.
+
+Either peer may send RPC requests to the other peer. Both acting as server and client at the same time.
+
+A common pattern is that one peer sends most of the RPC requests, while the other peer only occasionally sends requests back to the client to deliver notifications. This pattern is common in many applications because of how they are structured, not because of the JSON-RPC protocol itself or because of how this library implements it.
+
+## json_rpc's role
+
+`json_rpc` is a Nim library that implements the JSON-RPC protocol to easily send and receive RPC requests. It works on any transport (e.g. HTTP, Sockets, WebSocket). It is designed to automatically generate marshalling and parameter checking code based on the RPC parameter types.
```
arnetheduck: might be worth mentioning the 3 main layers here, ie "transports", "routing" and "dsl" where an application can choose to use any of them

docs/src/remote_targets.md
```diff
@@ -0,0 +1,51 @@
+# Adding Remote RPC Targets
```
arnetheduck: why not call it proxy in the title?

docs/src/remote_targets.md
```diff
@@ -0,0 +1,51 @@
...
+
+There are scenarios where users may need to add remote RPC targets to facilitate communication between two endpoints that have no direct RPC connection channel. Consider the following 3 endpoints:
+
+- client
+- server
```
arnetheduck: this framing is a bit odd - ie a "server" might reach out to other rpc services as part of its operation _or_ it might simply forward requests as they were given by the client - in the latter case, it's a proxy, in the former, it's an implementation detail - so I think this list needs to make up its mind ;)

docs/src/exceptions.md
```diff
@@ -0,0 +1,37 @@
...
+
+```nim
+{{#shiftinclude auto:../examples/http_server.nim:RpcTeaPot}}
+```
+
+## Client-side concerns
+
+An invocation of an RPC method may throw several exceptions back at the client. The base exception `JsonRpcError` can be used to catch all RPC exceptions.
+
+These are the exceptions which the client should be prepared to handle: `RpcTransportError`, `InvalidResponse`, `RequestDecodeError`, and `JsonRpcError`.
```
arnetheduck, line 29: oof, we really need to work towards having the `ApplicationError` be raised here for the non-reserved error codes

docs/examples/client_sigs_def.nim
```diff
@@ -0,0 +1 @@
+proc hello(input: string): string
```
arnetheduck, line 1: nim1 intentional?

docs/examples/http_server.nim
```diff
@@ -0,0 +1,67 @@
...
+  await srv.stop()
+  await srv.closeWait()
+  # ANCHOR_END: ServerDisconnect
+
+proc main() {.raises: [JsonRpcError].} =
+  let srv = startServer()
+  runForever()
+
+# Pass -d:jsonRpcExample to nim to run this
+when defined(jsonRpcExample):
```
arnetheduck: why not `when isMainModule`?

docs/examples/websocket_server.nim
```diff
@@ -0,0 +1,36 @@
...
+
+proc startServer*(): RpcWebSocketServer {.raises: [JsonRpcError].} =
+  # ANCHOR: ServerConnect
+  let srv = newRpcWebSocketServer("127.0.0.1", Port(0))
+  # ANCHOR_END: ServerConnect
+  srv.setupServer()
+  srv.start()
+  srv
+
+proc stopServer*(srv: RpcWebSocketServer) {.async.} =
```
arnetheduck: we typically want these to be `async: (raises: [])`, specially since the other functions do use raises annotations - ie shutting things down should always unconditionally work and should contiune closing even during attempted cancellations, ie errors are either returned for informational purposes or ignored

docs/examples/rpc_format.nim
```diff
@@ -0,0 +1,22 @@
...
+{.push raises: [], gcsafe.}
+
+import
+  json_serialization
+
+export
+  json_serialization
+
+# ANCHOR: FormatRpcConv
+createJsonFlavor RpcConv,
```
arnetheduck, line 13: in the flavor cookbook / explanation, it's probably useful to include a copy-pastable specialization of a readValue/WriteValue pair that shows how it can be used (even if it's covered in detail by json_ser docs, we want to have it accessible in this guide to show end-to-end how it works in json_rpc)

arnetheduck (approved):
nice! there's some polish to do and the guide highlights drives home the need for some additional polish here and there in the api - but otherwise, great progress

docs/examples/http_server.nim
nitely: so the CI can run the examples, but maybe it should just compile them without running them.

or avoid running the `*_server.nim` ones.

docs/src/connecting.md
nitely: it is, but it only shows the first tab content when printing the book...

docs/src/connecting.md
nitely: Mentioning the server can send requests, although that's not implemented yet does not seem very useful. I know it can send notifications, but the current implementation seem useful only in very niche cases (it sends it to all connected clients in that server instance, iirc), so that's not mentioned in the docs either.

json_rpc/private/server_handler_wrapper.nim
```diff
@@ -353,5 +354,8 @@ func wrapServerHandler*(
       # Avoid 'yield in expr not lowered' with an intermediate variable.
       # See: https://github.com/nim-lang/Nim/issues/17849
       `setup`
-      let handlerRes = `executeCall`
-      maybeWrapServerResult(`formatType`, handlerRes)
+      when `executeCallReturnType` is void:
+        `executeCall`
+      else:
+        let handlerRes = `executeCall`
+        maybeWrapServerResult(`formatType`, handlerRes)
```
nitely, line 361: #279

docs/src/connecting.md
arnetheduck: > useful only in very niche cases

oh? hm, that's unexpected - I thought it was used for subscription notifications in nimbus, ie the "client" sends a request to subscribe to some topic and then notifications are sent to the client for that topic, like server side events in http .. would be strange that they are sent to all clients. 

I guess this needs to be worked on, ie we should have two "peer factories" - one is the server that accepts incoming connections and the other does outgoing connections , but it's all a transport level distinction essentially and once they've negotiated, it's all the same.

nitely: maybe I misused it somehow?

https://github.com/status-im/nim-json-rpc/blob/fc317884984d02a995038e4a67654ad7b84886dc/json_rpc/server.nim#L96

```nim

import ./json_rpc/[rpcserver, rpcclient]

proc setupServer(srv: RpcServer) =
  srv.rpc(JrpcConv):
    proc notifier(): JsonString {.async.} =
      await srv.notify("notify", default(RequestParamsTx))
      JsonString("null")

createRpcSigsFromNim(RpcClient, JrpcConv):
  proc notifier(): JsonString

proc main {.async.} =
  var srv = newRpcWebSocketServer("127.0.0.1", Port(0))
  setupServer(srv)
  srv.start()

  var router1 = new RpcRouter
  router1[].rpc("notify") do() -> void:
    echo "NOTIFIED 1"

  var client1 = newRpcWebSocketClient(router = router1)
  await client1.connect("ws://" & $srv.localAddress())

  var router2 = new RpcRouter
  router2[].rpc("notify") do() -> void:
    echo "NOTIFIED 2"

  var client2 = newRpcWebSocketClient(router = router2)
  await client2.connect("ws://" & $srv.localAddress())

  discard await client1.notifier()
  await sleepAsync(3_000)

  await client1.close()
  await client2.close()
  srv.stop()
  await srv.closeWait()
  echo "ok"

waitFor main()

# NOTIFIED 1
# NOTIFIED 2
# ok
```

nitely: > ie the "client" sends a request to subscribe to some topic and then notifications are sent to the client for that topic, like server side events in http

In a SSE server you usually have something like `proc serverHandler(client: ClientCtx) = while true: client.sendSSEMessage( pubsub.getMessages(client.getTopics) )` which makes things easier. In Json-rpc you'd need to get the client instance somehow in the RPC handler and register it into some map of `topic -> connections`, I'd assume. Then every time the server gets a message for a topic (somehow), send it to all conns in that topic. But there is currently no way to get the client instance in a RPC handler to register it. It is also needed for the bidi general case.

nitely: > I thought it was used for subscription notifications in nimbus

for getting notifications in the client I guess it works (what I've seen it used for in web3). For *sending* them from the server there does not seem to be a easy way to send notifications/requests to the incoming client only.

PR #259, opened 2026-01-08
Fix #239; Allow custom JSON flavors for "client" code

nitely:
Fix #239

Changes:

- Add flavor as last param to client/server APIs.
- Set `JrpcConv` as default flavor for backward compat.
- Add `rpc(server, format, procList)` API to avoid repeating passing the server and flavor params in every rpc.

tested in nimbus-eth1/2, nim-web3; see linked PRs.

PR #264, opened 2026-02-24
Make async pragma required in rpc context API

nitely:
This makes async pragma explicit in RPC handlers and allows to define the raises list `{.async: (raises: []).}`. If the RPC is not async, the generated handler is non-async. All pragmas are copied as is into the generated handler in both cases.

Changes:

- RPC functions with async pragma generates async handlers.
- Non-async RPC functions no longer generates async handlers. The handler wrapper is still async; making the wrapper non-async would require adding a routes table for non-async callbacks, this can be done later.

When there are unlisted raises exceptions, the compile error does point to the right function.

No changes in the `srv.rpc("path") do(): ...` API. It still generates async handlers always, and ignores pragmas.

This is non-breaking given there is no release with the new API #259 yet

### DIFF
--- a/tests/test_router_rpc.nim
+++ b/tests/test_router_rpc.nim
@@ -66,6 +66,27 @@ server.rpc("returnJsonString") do(a, b, c: int) -> JsonString:
 server.rpc("serializedFN") do(a{.serializedFieldName: "result".}: int) -> int:
   return a
 
+server.rpc(JrpcConv):
+  proc rpcCtxAsync(s: string): string {.async.} =
+    await noCancel sleepAsync(0)
+    return "ret1 " & s
+
+  proc rpcCtxAsyncNoRaises(s: string): string {.async: (raises: []).} =
+    await noCancel sleepAsync(0)
+    return "ret2 " & s
+
+  proc rpcCtxAsyncWithRaises(s: string): string {.async: (raises: [ValueError]).} =
+    raise (ref ValueError)(msg: "err")
+
+  proc rpcCtxSync(s: string): string =
+    return "ret3 " & s
+
+  proc rpcCtxSyncNoRaises(s: string): string {.raises: [].} =
+    return "ret4 " & s
+
+  proc rpcCtxSyncWithRaises(s: string): string {.raises: [ValueError].} =
+    raise (ref ValueError)(msg: "err")
+
 func req(meth: string, params: string): string =
   """{"jsonrpc":"2.0", "method": """ &
     "\"" & meth & "\", \"params\": " & params & """, "id":0}"""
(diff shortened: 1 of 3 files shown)


tests/test_router_rpc.nim
```diff
@@ -66,6 +66,27 @@ server.rpc("returnJsonString") do(a, b, c: int) -> JsonString:
 server.rpc("serializedFN") do(a{.serializedFieldName: "result".}: int) -> int:
   return a
 
+server.rpc(JrpcConv):
+  proc rpcCtxAsync(s: string): string {.async.} =
```
arnetheduck, line 70: if we add {.async.} we also have to add `Future[...]` to return type .. hm .. I can't make up my mind whether this is a welcome change or not. maybe, since we're dealing with rpc, we can add `{.noasync.}` to _disable_ async instead .. ie that feels .. more in tune with this library maybe? thoughts?

nitely: eh, the reason I made this PR is because of ur comment here: https://github.com/status-im/nim-web3/pull/237#discussion_r2826437252

> Not really for this PR, but one thing that has always put me off a bit with json_rpc is the fact that these functions are implicitly async without this being visible in the signature or anywhere really

which I agree with. Adding `{.noasync.}` does not address this. Also, I think `Future[T]` return type is questionable, [yasync](https://github.com/yglukhov/yasync) doesn't have it and it's much nicer, it could be made optional in chronos, but it's a different discussion. I think it could be made optional here.

nitely: Also, copying the pragma into the generated handler makes async checked exceptions possible in rpc...

jangko: Some of nimbus-eth1 engine API relying on the fact that the rpc method is an async proc and make use of it to handle e.g. async block import. While the majority of the regular rpc methods do not really need async. It will be nice if the rpc macro can support both async and sync methods. So we can explicitly write async proc where it need to be and others are simply sync proc.

nitely: yes, I did updated my nimbus-eth1 PR to the changes in this PR https://github.com/status-im/nimbus-eth1/pull/3914/changes

nitely:
@jangko Jacek was not happy about this change, though. The alternative is to revert this, revert the relate PR uptades, and add `{.noasync.}` as an optimization later.

jangko:
Yeah, but doesn't this PR also support plain proc/sync proc without the need to add anything to the existing proc?
We still can live with those few async proc.

And I'm not happy if we add explicit `Future[...]` to the rpc methods, `{.async.}` pragma should enough implicitly tell the user that that particular rpc method is an async proc with return type wrapped by `Future[...]`.

Explicit `Future[...]` should be supported optionally. And `{.noasync.}` is weird, any plain rpc without `{.async.}` pragma should be treated as synchronous proc. I know it's a breaking changes, but also better than `{.noasync.}`, in real applications, adding `{.sync.}` probably less than adding `{.noasync.}`.

arnetheduck:
> Explicit Future[...] should be supported optionally. 

The underlying argument for explicit future is that this is what chronos itself does, so it's easier to remember one style and use it consistently both for regular async and rpc - one reason it's not optional there is that you can have `Future[Future[...]]`) making it difficult to auto-detect. It could be argued that it was a mistake to have the explicit future in chronos' async macro, but that ship has sailed.

Otherwise, it gets pretty odd - inside the rpc macro we would have `async` without Future and outside of it we have the same `async` keyword but with Future ..

So the way I see it we have two options:

* use the async keyword _and_ enforce `Future`
* use a different keyword (which can be `sync`, `noasync` or something like `rpcopts: (async: false)`

nitely:
> one reason it's not optional there is that you can have Future[Future[...]]) making it difficult to auto-detect.

I thought about that, and aside from it being very uncommon ([see search results in github](https://github.com/search?q=%22Future%5BFuture%5B%22+path%3A.nim&type=code)), I don't think it's an issue. Only non-Future types are rewritten to Future[T]. So `Future[T]` no matter what the T is (it could be a Future) would keep working as it works now (the change would be backward compatible). There would be an inconsistency in that if you want to return a Future (ex: after await) you need to write Future[Future[T]] instead of Future[T], yes, but that's pretty obvious if Future is optional.

Although, making it optional may not be a good idea if that will create bike shading in PRs.

Issue #261, opened 2026-02-19
RpcWebsocketClient raises InvalidResponse: Expected id 4, got 5


markspanbroek:
Performing several request in parallel to a hardhat server over websockets leads to `InvalidResponse` errors, complaining that the response id does not match the request id. This seems to be a regression in master (7434eaac677939984639b85dbe17afe557d6c45e) since version 0.5.4.

This bug can be reproduced in the context of nim-ethers here: https://github.com/durability-labs/nim-ethers/pull/12

arnetheduck:
Are the responses arriving out of order?

markspanbroek:
> Are the responses arriving out of order?

Yes, it seems that they are. I printed the responses [that come in from the websocket](https://github.com/status-im/nim-json-rpc/blob/7434eaac677939984639b85dbe17afe557d6c45e/json_rpc/clients/websocketclient.nim#L108) which gives this output:

```
{"jsonrpc":"2.0","id":1,"result":"0x4"}
{"jsonrpc":"2.0","id":2,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":3,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":4,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":5,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":6,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":7,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":8,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":10,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":9,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":11,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":12,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
{"jsonrpc":"2.0","id":13,"result":"0x000000000000000000000000000000000000000000000000000000000000000c"}
[...]

PR #266, opened 2026-04-08
Remove Nim 1.6 support

narimiran:


arnetheduck:
#261 needs fixing before we release 0.6

json_rpc.nimble
```diff
@@ -10,14 +10,14 @@
...
-version       = "0.5.4"
+version       = "0.6.0"
 author        = "Status Research & Development GmbH"
 description   = "Ethereum remote procedure calls"
 license       = "Apache License 2.0"
 skipDirs      = @["tests"]
 
 ### Dependencies
-requires "nim >= 1.6.0",
+requires "nim >= 2.0.8",
```
arnetheduck, line 20: > /home/runner/work/nim-json-rpc/nim-json-rpc/nimbledeps/pkgs2/websock-0.2.2-3c424661eff56c925b01e1cd1a911ff744e72962/websock/frame.nim(79, 9) Error: type mismatch: got <openArray[byte], MaskKey, int>

this is a bug that should be fixed in websock 0.2.2 - it's incorrectly claiming 1.6 support which makes nimble choose the wrong version here (it should have gracefully degraded to 0.2.1 for example) - changing the limit here does not actually fix the underlying issue.

tersec: - https://github.com/status-im/nim-websock/pull/190

narimiran:
Resolved by #269. Closing.

PR #267, opened 2026-04-10
avoid chunked encoding / data copies

arnetheduck:
* use a body writer to avoid copying data during request sends
* avoid chunked encoding in server - since the full response is being held in memory, there's no advantage of chunking it on the server side and we can use the simpler plain response

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ jobs:
   build:
     uses: status-im/nimbus-common-workflow/.github/workflows/common.yml@main
     with:
-      nimble-version: 6fe9d817a70a14fa57022a9cca46eb443ee5a040
       test-command: |
         nimble setup -l
         nimble test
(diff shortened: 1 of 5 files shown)


jangko:
The downside of this PR is we cannot test if the _client_ actually can parse chunked encoding anymore. In the first place the chunked encoding support is added because in hive or somewhere else the client is expected to support chunked encoding. But I agree chunked encoding is disabled in the server. Probably we can support both mode in the server so we still can test if the client is still working properly. And the default mode for the server is plain mode.

arnetheduck:
>  actually can parse chunked encoding

This is handled (and tested) in chronos now - basically, chunked encoding should be transparent to a "consumer" like json-rpc - aka json-rpc isn't the right place to test chunked encoding I think and testing elsewhere has improved sufficiently to not have to test it here any more.

arnetheduck:
ok, thought about it and added a simple hard-coded chunked encoding test for such a server

PR #270, opened 2026-04-13
Fix #261 regression; Support unordered responses again

### LINKED ISSUE #261: RpcWebsocketClient raises InvalidResponse: Expected id 4, got 5

nitely:
Fix #261

Changes:

- Support unordered responses
- Adds `BidiMessage` to parse request/response messages a single time.

Changes wrt v0.5.4:

- On ambiguous messages there is no way to know whether the message is a request or a response; hence we send an invalid request error response and terminate the connection (even if the message was not a request, sending a response and terminating should be fine). In v0.5.4 there was no bidirectional communication so it was always known the message was a request, so the connection could be kept alive safely.
- Unknown response IDs are ignored instead of failing all pending requests. Unknown IDs: int id not in the pending requests table, or id of type string. Receiving a null id fails all pending requests (one of the requests failed but we cannot know which one), and terminates the connection; but I think this could be changed to keep going safely given next unknown id (likely removed from pending requests) will be ignored.
- Pending request Future is removed from the table if the user cancels the request (ie: adds a timeout). Old behavior was to keep the future in the table forever, or until the table got cleared by some failing request.
- Batch requests are added to the pending request table, and last id within the batch is used to match the entry. This enables waiting on multiple batch requests at once.


json_rpc/client.nim
```diff
@@ -119,50 +133,77 @@ proc callOnProcessMessage*(
...
+  of rbkSingle:
+    singleResponseId(batch.single)
+  var fut: ResponseFut
+  if client.pendingRequests.pop(id, fut):
+    if not fut.finished():
+      fut.complete(batch)
+    else:
+      debug "Future already finished, dropping", state = fut.state()
+  else:
+    debug "Pending request id not found", id = id
```
nitely, line 168: v0.5.4 used to fail all pending requests and continue processing, which caused next response to contain an unknown id, which again caused failing all pending requests and so on. Ignoring unknown IDs seems better, but alternatively we could fail all requests and terminate the connection.

Discussed in https://github.com/nitely/nim-json-rpc/pull/4#discussion_r3060242732

> [arnetheduck](https://github.com/arnetheduck) [last week](https://github.com/nitely/nim-json-rpc/pull/4#discussion_r3060242732)
> 
> this is only needed when relying on order - ie if a message arrives that is out-of-order it was not possible to do anything better ..
> 
> in the version with a table, we can simply ignore "unknown" responses (at the risk of seeing the table of expected requests grow indefinitely ..

> 
> [nitely](https://github.com/nitely) [4 days ago](https://github.com/nitely/nim-json-rpc/pull/4#discussion_r3068615580)
> it makes sense to ignore missing ids, since the user should add a timeout to the method call anyway, since there is no guarantee the server will ever respond. The behavior differs from the HTTP transport, which will raise an error if the response is missing the id, though.
> 
> The v0.5.4 behavior of clearing all pendings is really bad, though, as it may generate never ending clearing of pending futures (since the server will send responses to requests already cleared, and so raise the error of unknown response and clear all current pending again, and so on).
> 
> I've changed the code to ignore responses with missing ids.
> 
> The alternative is to clear the pending requests, send invalid request error, and terminate the connection... that's more in line with the old v0.5.4 behavior

json_rpc/client.nim
```diff
@@ -119,50 +133,77 @@ proc callOnProcessMessage*(
...
+    resp.id.num
+  of riString:
+    int.low
+  of riNull:
+    case resp.kind
+    of rkResult:
+      int.low
+    of rkError:
+      # likely an invalid request error
+      raise (ref JsonRpcError)(msg: JrpcSys.encode(resp.error))
```
nitely, line 148: This will fail all pending requests and terminate the connection. But I think it's safe to fail the pending requests and keep going (the pending responses will contain an unknown id and get ignored, although they will be pointlessly processed). It'd need to throw some other error type, though.

json_rpc/client.nim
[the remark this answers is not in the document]
nitely: it seems to be a leftover from the bidirectional refactor. The original method seems to be overloaded in the wild here https://github.com/durability-labs/nim-ethers/blob/9004f7c3abc537bd7fb7fbf050f8e5d070f73821/ethers/providers/jsonrpc/rpclimits.nim#L39 but it already had a fix https://github.com/durability-labs/nim-ethers/pull/13/changes (later reverted to use old v0.5.4).

so removing it should be fine. There is no other relevant match in github https://github.com/search?q=%22callBatch%22%20path%3A.nim&type=code and the previous callBatch accepted a different type anyway.

json_rpc/client.nim
```diff
@@ -248,40 +285,6 @@ proc call*(
 ): Future[JsonString] {.async: (raises: [CancelledError, JsonRpcError], raw: true).} =
   client.call(name, paramsTx(params, JrpcConv))
 
-proc callBatch*(
```
nitely, line 251: there is no test or usage for this proc, and there is no way to create a `RequestTx` without importing from private it seems. It can probably be removed, not sure what was the idea for this api.

it seems to be a leftover from the bidirectional refactor. The original method seems to be overloaded in the wild here https://github.com/durability-labs/nim-ethers/blob/9004f7c3abc537bd7fb7fbf050f8e5d070f73821/ethers/providers/jsonrpc/rpclimits.nim#L39 but it already had a fix https://github.com/durability-labs/nim-ethers/pull/13/changes (later reverted to use old v0.5.4).

so removing it should be fine. There is no other relevant match in github https://github.com/search?q=%22callBatch%22%20path%3A.nim&type=code and the previous callBatch accepted a different type anyway.

json_rpc/client.nim
arnetheduck: if someone is sending us junk there's scant reason to keep connected I suspect .. ie the connection goes into UB effectively so even if there's an `id` match, it could just as well be wrong.

nitely: this occurs when the id is null; ex: if the server failed to parse the request.

then we should terminate the conn if the id is a string too, right?

arnetheduck: `id` can be a `string` or `number`: https://www.jsonrpc.org/specification#request_object

nitely: or `null`, and the conn is terminated if null, why not if it's string as well? the spec also says the server must return the same received id and we always send a number not a string.

arnetheduck: right .. `null` is a notification and I thought we handled that somewhere - if this code removes that handling, it needs fixing, ie we should then run the rpc but not respond ..

arnetheduck: > we always send a number not a string

other rpc clients might use a string and they expect it to be mirrored back to them .. `JsonString` might be appropriate

nitely: we are talking different things. This code handles responses, `null` id in a response is a request error (or some other error). Notifications don't get responses at all. Requests ~~with id `null`~~ without id are notifications that don't expect responses.

> other rpc clients might use a string and they expect it to be mirrored back to them .. JsonString might be appropriate

There are probably no tests for it, but the server does reply with a string id if it got a request with a string id.

arnetheduck: aaah ok .. yeah, if this is only for responses to things we sent, disconnecting on everything we couldn't have sent seems fine

PR #275, opened 2026-05-04
Fix #274: Invalid `meth` implementation.

### LINKED ISSUE #274: Bug in `meth` function: can't share the same code for RequestRx and RequestRx2
In this function, we attempt to get `method` from rx and return it as a string:

https://github.com/status-im/nim-json-rpc/blob/master/json_rpc/private/jrpc_sys.nim#L216

But depending on whether it's a RequestRx or RequestRx2, `method` is either an `Opt[string]` or a `string`.

moigagoo:
Split `meth` into to funcs: one for RequestRx and one for RequestRx2.

### DIFF
--- a/json_rpc/private/jrpc_sys.nim
+++ b/json_rpc/private/jrpc_sys.nim
@@ -213,8 +213,8 @@ func `==`*(a, b: RequestId): bool =
   of riString: a.str == b.str
   of riNull: true
 
-func meth*(rx: RequestRx): string =
-  rx.`method`.get
+func meth*(rx: RequestRx): Opt[string] =
+  rx.`method`
 
 func meth*(rx: RequestRx2): string =
   rx.`method`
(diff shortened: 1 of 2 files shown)


json_rpc/private/jrpc_sys.nim
```diff
@@ -213,7 +213,10 @@ func `==`*(a, b: RequestId): bool =
   of riString: a.str == b.str
   of riNull: true
 
-func meth*(rx: RequestRx | RequestRx2): string =
+func meth*(rx: RequestRx): string =
+  rx.`method`.get
```
nitely: I think this should return Opt[string] given it's an alias for method, and it may raise a defect if isNone.

nitely: ye, it used to return Opt in v0.5.4 https://github.com/status-im/nim-json-rpc/blob/b6e40a776fa2d00b97a9366761fb7da18f31ae5c/json_rpc/private/jrpc_sys.nim#L179

moigagoo: Updated.

PR #276, opened 2026-05-05
Terminate conn on invalid response `id` & don't send a response for it

nitely:
Changes:

- Terminate bidi conn on invalid response `id`. As discussed in https://github.com/status-im/nim-json-rpc/pull/270#discussion_r3090028248
- Avoid sending an "invalid request" response on invalid response `id`. It should only be done on ambiguous messages before terminating.

### DIFF
--- a/tests/test_bidirectional.nim
+++ b/tests/test_bidirectional.nim
@@ -7,6 +7,8 @@
 # This file may not be copied, modified, or distributed except according to
 # those terms.
 
+{.push raises: [], gcsafe.}
+
 import
   chronos/unittest2/asynctests,
   stew/byteutils,
@@ -152,8 +154,7 @@ template allTests(client: untyped) =
 
 suite "Test bidirectional socket server/client":
   setup:
-    # XXX Framing.lengthHeaderBE32()
-    const framing = Framing.newLine()
+    const framing = Framing.lengthHeaderBE32()
     var srv = newRpcSocketServer(["127.0.0.1:0"], framing = framing)
     var client = newRpcSocketClient(framing = framing)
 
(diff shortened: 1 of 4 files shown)


tests/test_bidirectional.nim
```diff
@@ -60,104 +79,62 @@ template allTests(client: untyped) =
...
-    # following requests still work
-    let r1 = waitFor client.rets("foobar")
-    check r1 == "ret foobar"
+  test "Sending an ambiguous batch message terminates the connection":
+    const req = """[{"foo": "boo"}]"""
+    const expected = """{"code":-32600,"message":"',' expected"}"""
+    checkInvalidRequest(client, req, expected)
+
+  test "Sending a null id terminates the connection":
+    # note this terminates the connection when receiving the null id response
```
arnetheduck, line 124: https://www.jsonrpc.org/specification#notification

> Notifications are not confirmable by definition, since they do not have a Response object to be returned. As such, the Client would not be aware of any errors (like e.g. "Invalid params","Internal error").

so .. we should not disconnect in this case since we should not be sending the error?

nitely: id null is not a notification, apparently. Notifications don't have the id field at all.

https://www.jsonrpc.org/specification#id1

> The use of Null as a value for the id member in a Request object is discouraged, because this specification uses a value of Null for Responses with an unknown id. Also, because JSON-RPC 1.0 uses an id value of Null for Notifications this could cause confusion in handling.

but this library does not support json-rpc 1.0, it fails if the jsonrpc field value is not 2.0

and also https://www.jsonrpc.org/specification#notification

> A Notification is a Request object without an "id" member.

it's also mentioned in the request id description https://www.jsonrpc.org/specification#request_object id null is a valid value and "the Server MUST reply with the same value in the Response object if included."

arnetheduck: right .. thanks for the clarification!

arnetheduck (approved):

PR #282, opened 2026-07-01
Restrict client raises to `JsonRpcError`

nitely:
Changes:

- Annotate client dsl wrapper with `raises: [CancelledError, JsonRpcError]` to avoid fallback to `CatchableError`.
- Export `errors` in client due to dsl requiring `JsonRpcError`
- Add `ResultDecodeError` to raise it when failing to decode the response result.
- Use `setup` and `teardown` in `test_callsigs`

### DIFF
--- a/json_rpc/private/client_handler_wrapper.nim
+++ b/json_rpc/private/client_handler_wrapper.nim
@@ -20,7 +20,8 @@ func createRpcProc(procName, parameters, callBody: NimNode): NimNode =
   for p in parameters: paramList.add(p)
 
   # build proc
-  result = newProc(procName, paramList, callBody)
+  let pragmas = quote do: {.async: (raw: true, raises: [CancelledError, JsonRpcError]).}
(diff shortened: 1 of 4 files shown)


json_rpc/private/client_handler_wrapper.nim
```diff
@@ -54,9 +54,15 @@ template maybeUnwrapClientResult*(client, meth, reqParams, returnType, formatTyp
...
     client.call(meth, reqParams)
   else:
-    proc complete(f: auto): Future[returnType] {.async.} =
+    proc complete(f: auto): Future[returnType] {.async: (raises: [CancelledError, JsonRpcError]).} =
       let res = await f
-      decode(formatType, res.string, returnType)
+      try:
+        decode(formatType, res.string, returnType)
+      except SerializationError as exc:
+        raise (ref JsonRpcError)(
```
nitely: this should be a `ResultDecodeError` with a `result` field set to `res` (JsonString)

PR #287, opened 2026-08-07
Fix #262; Client side response error should contain the JSON error fields

### LINKED ISSUE #262: Client side response error should contain the JSON error fields
When a method invocation fails with an application or server error, [the raised error](https://github.com/status-im/nim-json-rpc/blob/7434eaac677939984639b85dbe17afe557d6c45e/json_rpc/client.nim#L93) is a `JsonRpcError` with the response error object encoded in the `msg` field. There is no easy way to decode this object, nor check whether the msg contains a response error. Given all errors inherit from `JsonRpcError`, the msg could be anything else.

A `JsonRpcResponseError` which contains the decoded fields should be raised instead. Then one could simply check `err.code` for special error handling for example. The msg would still be the encoded response error for logging purposes.

nitely:
Fix #262

Changes:

- Raise proper server response exception per code: `ApplicationError`, `ServerError`, etc. Breaking change because the payload is no longer stored as the exception msg. Also, if the user mistakenly raises an `ApplicationError` with a server error code, client gets a ServerError instead of ApplicationError.
- Forward server errors in proxy server. Breaking change, but needed.
- Update exception docs and examples
- `ApplicationError` renamed to `RpcApplicationError` for consistency; but there is an alias for `ApplicationError = RpcApplicationError`. Also potentially breaking; it could clash with some other object named the same, as well as all the new exceptions.


nitely:
Pre-existent issues I noticed and I didn't fix (they would be breaking changes):

- "invalid params" is returned as a server error (-32000) instead of invalid params (-32602); ~~not sure if worth fixing.~~
- ApplicationError is propagated from client calls done within an RPC method; which could be confusing. Server errors as well, but that can be a bit less confusing as the original msg is added to data, and the msg looks ok. In theory this may leak internal/secret (or third-party) error messages and it cannot be prevented with raises annotations (if the method already raise its own ApplicationError). The "fix" would be adding an enum field of local/remote (local default value, remote set when raising from json-rpc internals) values to the exception; set a fixed message for remote, and just propagate local.

arnetheduck:
cc @kdeme @bhartnett care to take a look at what this will do to our downstream usage?

bhartnett:
> cc @kdeme @bhartnett care to take a look at what this will do to our downstream usage?

These changes have minimal impact on nimbus-eth1. Looks like this doesn't change the rpc server behavior, just the rpc client and so just minor changes would be needed:
1. Tests that use the rpc client would need to be updated to handle the change. 
2. The portal rpc client (used in the portal bridge and nimbus el portal network block fetching) would need to update the error handling in `toPortalRpcError`. Nothing major or blocking.

nitely:
I've added `proc toJsonError*(e: ref RpcAnyServerError): string` which returns the encoded json error if needed.

I've made tests pass in https://github.com/nitely/nimbus-eth1/pull/10 ; I can send a patch to nimbus-eth1 with these changes later.

I grepped `JsonRpcError` and `ApplicationError` in nimbus and the change seems ok. One thing I noticed is the [engine API codes](https://github.com/status-im/nim-web3/blob/a062a1fd87084a936b54d8dc37bcc33241705801/web3/engine_api_types.nim#L315) reuse the json-rpc spec defined error codes; so you get server errors for those instead of application error.

[Here is another instance](https://github.com/status-im/nimbus-eth1/blob/f787a74bb38e8bf4741719ebf853811d06159d6d/execution_chain/rpc/common.nim#L112), in this case you get `RpcServerInvalidParamsError` client side...

Maybe there should be one single response error type instead... but then `ApplicationError` would be an alias for every response error code; although it could be deprecated in favor of a more general `RpcResponseError`.

json_rpc/errors.nim
```diff
@@ -46,12 +46,66 @@ type
     ## raised when failing to decode the response result
     result*: JsonString
 
-  ApplicationError* = object of JsonRpcError
+  RpcAnyServerError* = object of JsonRpcError
+    ## Base error for all server error responses
+    code*: int
+    data*: results.Opt[JsonString]
```
arnetheduck: do we need to care about empty vs not present here?

json_rpc/errors.nim
```diff
@@ -46,12 +46,66 @@ type
     ## raised when failing to decode the response result
     result*: JsonString
 
-  ApplicationError* = object of JsonRpcError
+  RpcAnyServerError* = object of JsonRpcError
```
arnetheduck:
```suggestion
  RpcResponseError* = object of JsonRpcError
```

"server" feels odd here since it also covers application errors which are not problems with the json-rpc protocol/server itself .. also, "any"` is always implied with a hierarchy so it feels superfluous.

json_rpc/errors.nim
```diff
@@ -46,12 +46,66 @@ type
...
-  InvalidRequest* {.deprecated: "ApplicationError".} = ApplicationError
+  InvalidRequest* {.deprecated: "RpcApplicationError".} = RpcApplicationError
+
+  # Aliases
+  ApplicationError* = RpcApplicationError
+
+# https://www.jsonrpc.org/specification#error_object
+proc init*(
+  T: type (ref RpcAnyServerError),
+  code: int,
```
arnetheduck, line 90:
```suggestion
proc new*(
  T: type (RpcAnyServerError),
```

maybe more convenient to spell, with `new` introducing `ref` to `T`

arnetheduck:
> "invalid params" is returned as a server error (-32000) instead of invalid params (-32602); not sure if worth fixing.

Yes, let's fix this while we're at it, specially since we use -32000 as a catch-all for exceptions from "user code" - better keep things cleanly separated.

> ApplicationError is propagated from client calls done within an RPC method; which could be confusing. 

Yes, good point - this is a general problem with exceptions and the main reason why we broadly avoid them (ie there's no way to _force_ an explicit call site annotation) - that ship has sailed, but I've often wanted to add a language feature for forcing callsite annotations (which unlike `raises` also enforces them.

> adding an enum field of local/remote

This sounds reasonable.

> internal

There's a similar argument to be made for `CatchableError` in general, that we should not propagate the "internal" error and instead write a generic error message. Here's what we can do:
* add a toggle to the server / router / somewhere which enables error details to be propagated - maybe even with a per-rpc override though this might be excessive - the important thing is to allow it to be configured per "service"
* also set the default with a compile-time option, defaulting to _no propagation_ - gives an easy place to enable it during testing (ie in integration tests etc)

This way, by default, all exception information _except_ the explicit user-created `RpcApplicationError` are replaced by generic messages that don't leak internals.

> One thing I noticed is the 

Yeah I consider this a spec bug but here we are - nothing the library can do now. 

We discussed this briefly and one approach is to let the application raise a `RpcApplicationError` with the code set to one of the "internal" codes. 

In other words, pursuant to the above discussion about errors, we would allow implementations to raise `RpcApplicationError` with any error code they like - these would get translated to json-rpc errors with code etc - _all other_ exceptions, including the general `RpcResponseError` would be mapped to a generic error. It would mean that a client catching these will have to deal with buggy protocols but I think this is a good-enough tradeoff.

> deprecated in favor of a more general 

I'm not entirely opposed to this but I don't have a strong opinion really .. I think when raising, it's good to keep the types apart so we could keep `RpcApplicationError` on top of `RpcResponseError` as a single "magic" type for the user code to raise and for the client code to catch, and leave it there - this way, clients can catch application errors or "any" response errors .. engine api would use the latter strategy while a less buggy protocol could split its catch handlers. At the end of the day, it's either a list of `except` clauses or an `except` + `case` in the handler ..

arnetheduck:
this reminds me of a trick we've used sometimes where we have `{.requiresInit.}` on the type and a private field so the error can only be constructed from within a specific module but still the type can be public (with a public getter).. this would be a way to keep "user code" from raising the "internal" error types. _Most of the time_ I think it's enough to document with a comment that user code should not raise these errors but that's one way to enforce it as well.

json_rpc/errors.nim
```diff
@@ -46,12 +46,66 @@ type
...
     ## be outside the range of -32768 to -32000. A custom JSON data object may
     ## be provided.
-    code*: int
-    data*: results.Opt[JsonString]
 
-  InvalidRequest* {.deprecated: "ApplicationError".} = ApplicationError
+  InvalidRequest* {.deprecated: "RpcApplicationError".} = RpcApplicationError
+
+  # Aliases
+  ApplicationError* = RpcApplicationError
```
arnetheduck:
```suggestion
  ApplicationError* {.deprecated.} = RpcApplicationError
```

one or the other but not both

nitely:
> I'm not entirely opposed to this but I don't have a strong opinion really ..

we could go further and have two distinct error types for local and remote errors. The method handler would raise `RpcLocalResponseError` (alias `ApplicationError = RpcLocalResponseError`) and the client would get `RpcRemoteResponseError`. It would make impossible to re-raise client call errors from an RPC method with `raises: [ApplicationError]`. Although, I think in practice that's not an issue, because the user needs to handle JsonRpcError in the client call and actively re-raise the `ApplicationError` there, and then if you propagate this from the rpc handler, it does not seem like surprising behavior. *But* it also removes the need for the local/remote enum exception field, and makes the distinction more obvious/visible.

It is also what `vs-streamjsonrpc` seems to do with its [LocalRpcException](https://microsoft.github.io/vs-streamjsonrpc/api/StreamJsonRpc.LocalRpcException.html) and [RemoteInvocationException](https://microsoft.github.io/vs-streamjsonrpc/api/StreamJsonRpc.RemoteInvocationException.html) described in [their docs](https://microsoft.github.io/vs-streamjsonrpc/docs/exceptions.html#server-side-concerns).

> At the end of the day, it's either a list of except clauses or an except + case in the handler ..

yes, this design would require checking the code which is a bit more cumbersome. Although, we could have specific `RpcRemoteMethodNotFoundError = object of RpcRemoteResponseError`, etc without it being inconsistent with what the rpc method throws... in other words there would be no "ApplicationError" to catch, but the opposite: `RpcRemoteResponseError` catches every code, and then `RpcRemoteXXXError` is used for the spec defined errors; this solves the inconsistency of ApplicationError can contain any code, but then a different error is raised which cannot be caught with ApplicationError. It can be done this way with the enum field approach as well. The ApplicationError must be deprecated in favor of RpcRemoteResponseError (or RpcResponseError with the enum) then.

json_rpc/errors.nim
nitely: not sure, but I think the Opt can be dropped given an empty JsonString means not present.

json_rpc/errors.nim
nitely: I think this should be dreprecated and it should be an alias for `RpcResponseError` given it can contain any code, and users are using it as RpcResponseError.

RpcApplicationError can be removed to avoid confusion.

nitely: changed to `ApplicationError* {.deprecated: "RpcResponseError".} = RpcResponseError` instead.

json_rpc/errors.nim
nitely: seems pretty breaking; but it seems only nimbus-eth1 uses ApplicationError data field https://github.com/search?q=%22ApplicationError%22+path%3A.nim&type=code

nitely:
@arnetheduck so, do we go with the shared client/server error type with enum field or the `vs-streamjsonrpc` Local/Remote errors?

either way *I think* it could be changed later without it being a breaking change (but with lots of deprecations, though).

json_rpc/errors.nim
arnetheduck: we can leave the old AE type in place and deprecate it (and have a compat shim in the handler except block)

arnetheduck:
Hm, maybe indeed the remote / local split works well, though I'd be careful to not make the naming _excessive_ since for most applications, this problem won't appear on their radar - `RpcRemoteMethodNotFoundError` for example is a really bad name - `R` means remote so this is a `remote-procedure-call-remote-method-not-found-error` which is nuts.

As far as I can tell, the distinction mostly makes sense for ApplicationError, where there's a local and remote version (and maybe the local version doesn't need a prefix, or vice versa) - as long as these two have logically consistent names, I think we should be fine.

nitely:
Remote/Local only make sense with the prefix. Otherwise it's confusing. Having `RpcMethodNotFoundError` why can't I raise that from the server? why do I have to raise `RpcLocalMethodNotFoundError` instead? ok, most likely `RpcLocalMethodNotFoundError` should not exist. You still have `RpcRemoteResponseError` and `RpcLocalResponseError` with the same issue.

> As far as I can tell, the distinction mostly makes sense for ApplicationError,

But ApplicationError is not being used as ApplicationError, it's being used as RpcResponseError. Even if there would be a `RpcResponseError` in HEAD one can still use `ApplicationError` to return any code. So as a server error it does not make sense; it only make sense for catching it client side.

I think I'm going with the enum field.

PR #289, opened 2026-08-08
Fix #285; Do not swallow cancellations in socket client

### LINKED ISSUE #285: Do not swallow cancellations in socket client
#284 needs to be done for `socketclient`, and also review the rest of CatchableError (see if they really need to be CatchableError); probably catch CancelledError in each for consistency (even if not needed; ex: processMessages must break instead of reraise) or add comment.

nitely:
Fix #285

Unlikely to be breaking given CancelledError is usually propagated anyway (in await fut, instead of the send call).

### DIFF
--- a/json_rpc/clients/socketclient.nim
+++ b/json_rpc/clients/socketclient.nim
@@ -211,7 +211,7 @@ method request(
   client.withPendingFut(fut, id):
     try:
       await client.framing.sendMsg(client.transport, reqData)
-    except CatchableError as exc:
+    except TransportError as exc:
       raise (ref RpcPostError)(msg: exc.msg, parent: exc)
 
     await fut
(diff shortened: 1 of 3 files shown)


nitely:
There is one cancellation swallow left in the [process messages loop](https://github.com/status-im/nim-json-rpc/blob/9ac5306745217accd54326b37c0f7dc5b77d843a/json_rpc/clients/websocketclient.nim#L125) (for the socket transport also). When closing the connection, it fails pending requests with a transport error instead of a cancelled error. I would expect the cancelled error, but maybe this was intended?

Although, the change is potentially quite breaking compared to the one in this PR; if closing while there are pending requests, it must be done from some other future indirectly, unlike this PR change which requires the cancel to be done in the call future itself.

